PR #58, opened 2022-03-01
Integrate azcore

magodo:
This PR replaces the underlying transport layer `autorest-go` with `azcore`. This automatically replaced the underlying auth package `go-azure-helper`/`hamilton` with `azidentity`/`msal-go`.

## Notable Changes

There are multiple changes introduced during the integration, which are listed below.

### Transport Layer

1. The Track1 SDK retry on 429 error is disabled. For track2 SDK, as is stated in the [CHANGELOG](https://github.com/Azure/azure-sdk-for-go/blob/ade5ec6551ed064938eefb8ce1eaaf753c03642c/sdk/azcore/CHANGELOG.md#features-added-42), since v0.4.0, 429 won’t be retried. However, it is added back in commit: [7b3181265a8da567cef05b17ac7decc6bbab27e0](https://github.com/azure/azure-sdk-for-go/commit/7b3181265a8da567cef05b17ac7decc6bbab27e0). Though we can change remove the 429 from the retry list, but there is no good reason to do so. So keep it as is.

### Auth Layer

1. For the SP Cert auth, the go-azure-helper can supports the cert file with/without password, while the track2 only [supports cert file without password](https://github.com/Azure/azure-sdk-for-go/blob/ade5ec6551ed064938eefb8ce1eaaf753c03642c/sdk/azidentity/environment_credential.go#L40-L41). https://github.com/Azure/azure-sdk-for-go/pull/17099 is submit to add that support.
2. For the MSI auth, the go-azure-helper allows users to specify the `msi_endpooint`, while the track2 allows users to specify [the ID of the managed identity](https://github.com/Azure/azure-sdk-for-go/blob/a32823ace3910332da4fb81d57b87b1866661bc9/sdk/azidentity/managed_identity_credential.go#L59-L62), which is different. Since in most cases, `msi_endpoint` is not needed to specify, so we can leave this feature absent for now.
3. Multitenant is currently not working for track2 SDK: https://github.com/Azure/azure-sdk-for-go/issues/17159

### Other feature

1. The `metadata_host` can be used to set a customized metadata host, which can return customized endpoint for authorization or/and responding ARM API.


internal/provider/provider.go
```diff
@@ -101,20 +108,28 @@ func azureProvider() *schema.Provider {
 				Description: "The Client Secret which should be used. For use When authenticating as a Service Principal using a Client Secret.",
 			},
 
-			// Managed Service Identity specific fields
-			"use_msi": {
+			"skip_provider_registration": {
```
ms-henglu, line 111: We don't support provider registration, so no need to expose this.

internal/provider/provider.go
```diff
@@ -138,49 +153,48 @@ func providerConfigure(p *schema.Provider) schema.ConfigureContextFunc {
...
+			return nil, diag.Errorf("failed to obtain a credential: %v", err)
 		}
 
-		clientBuilder := clients.ClientBuilder{
-			AuthConfig:       config,
-			TerraformVersion: terraformVersion,
+		copt := &clients.Option{
+			SubscriptionId:       d.Get("subscription_id").(string),
+			Cred:                 cred,
+			AuxiliaryTenantIDs:   auxTenants,
```
ms-henglu, line 190: I thought `auxiliary_tenant_ids` is not supported?

ms-henglu (changes requested):

internal/provider/provider.go
magodo: This is supported in the track2 SDK: https://github.com/Azure/terraform-provider-azapi/blob/681dac0a2c6b5fdfe0b6d936a5a259e28c88165f/vendor/github.com/Azure/azure-sdk-for-go/sdk/azcore/arm/client_options.go#L30-L31

ms-henglu: Can we add `skip_provider_registration` to the doc?

internal/services/azurerm_generic_resource.go
```diff
@@ -202,15 +204,14 @@ func resourceAzureGenericResourceCreateUpdate(d *schema.ResourceData, meta inter
...
-				return fmt.Errorf("checking for presence of existing %s: %+v", id, err)
-			}
-		}
-		if len(utils.GetId(existing)) > 0 {
+		_, _, err := client.Get(ctx, id.AzureResourceId, id.ApiVersion)
+		if err == nil {
 			return tf.ImportAsExistsError("azapi_resource", id.ID())
 		}
+		var responseErr *azcore.ResponseError
+		if !errors.As(err, &responseErr) || responseErr.StatusCode != http.StatusNotFound {
```
magodo, line 212: Extract a `ResponseIs404` function..

ms-henglu:
LGTM 👍

Issue #73, opened 2022-03-31
Incorrect `attributes.id` property stored in state when specifying `parent_id = "/"`


matt-FFFFFF:
Hi,

I am trying to test deploying resources at tenant scope `/`. In this case an empty ARM template.

This resource declaration actually creates the resource, but the `resourceAzApiResourceRead` func returns an error after creation, and subsequent `terraform plan` commands also fail.

## Error

```
azapi_resource.tenant_arm_deployment: Refreshing state... [id=//providers/Microsoft.Resources/deployments/tenantdeployment1234]
╷
│ Error: `resource_id` and `type` are not matched
│
│   with azapi_resource.tenant_arm_deployment,
│   on main.tf line 13, in resource "azapi_resource" "tenant_arm_deployment":
│   13: resource "azapi_resource" "tenant_arm_deployment" {
│
```

You can see the issue in that the resource id begins with two slashes - `//`, instead of one. If I manually edit the state to remove the extra `/` then it works as expected.

## Problematic resource declaration

```terraform
resource "azapi_resource" "tenant_arm_deployment" {
  type      = "Microsoft.Resources/deployments@2021-04-01"
  name      = "tenantdeployment1234"
  location  = "northeurope"
  parent_id = "/"

  body = jsonencode({
    properties = {
      mode = "Incremental"
      template = {
        "$schema"      = "https://schema.management.azure.com/schemas/2019-08-01/tenantDeploymentTemplate.json#"
        contentVersion = "1.0.0.0"
        outputs        = {}
        resources      = []
        variables      = {}
        parameters     = {}
      }
    }
  })
}
```

## Other information

This only happens if `parent_id = "/"`, if I set `parent_id = ""` then it works.

ms-henglu:
Hi @matt-FFFFFF ,

Thank you so much for taking your time to report this issue!!

I would suggest use empty string "" as `parent_id` for these tenant scope resources. And I'll raise a PR to prevent users setting `/` as the `parent_id`.

ms-henglu:
Hi @matt-FFFFFF ,

Thank you so much for taking your time to report this issue!!

I've raised a PR to fix this issue, and will add a validation to prevent user set `parent_id = ""`, which means "/" is the only allowed value.

matt-FFFFFF:
Thanks @ms-henglu for the quick fix!

Confirmed working 🥳

Issue #72, opened 2022-03-31
`parent_id` value for creating `Microsoft.Management/managementGroups/settings` resource


matt-FFFFFF:
Hi,

I'm trying to create a management group hierarchy settings object as per this but cannot work out the correct `parent_id`. The API docs are here:

https://docs.microsoft.com/en-us/rest/api/managementgroups/hierarchy-settings/create-or-update

When using the `az rest`, the `PUT` url is as follows, this is accepted by the RP:

```bash
az rest --method PUT --url /providers/Microsoft.Management/managementGroups/{groupId}/settings/default?api-version=2021-04-01 --body @hierarchySettings.json
```

When I try and implement this using the `azapi` provider I am struggling to work out how to do it.

I believe the below should be correct:

```terraform
resource "azapi_resource" "hierarchy_settings" {
  type      = "Microsoft.Management/managementGroups/settings@2021-04-01"
  name      = "default"
  parent_id = "/providers/Microsoft.Management/managementGroups/bb4c9e42-b982-4573-8741-cb71d8b92af4/settings/default"
  body = jsonencode({
    properties = {
      defaultManagementGroup               = "/providers/Microsoft.Management/managementGroups/demo-corp"
      requireAuthorizationForGroupCreation = true
    }
  })
}
```

But I get the error:

```
Error: `parent_id` is invalid, expect id of resource whose scope is [Tenant], but got scope ManagementGroup`
```

I have been trying various different `parent_id` values and cannot make it work. Any help appreciated 🆘

matt-FFFFFF:
If I specify `parent_id = ""`

Then the plan works but the apply fails with this:

```
╷
│ Error: `parent_id` is invalid, expect id of `Microsoft.Management/managementGroups`
│
│   with azapi_resource.hierarchy_settings,
│   on main.tf line 1, in resource "azapi_resource" "hierarchy_settings":
│    1: resource "azapi_resource" "hierarchy_settings" {
```

ms-henglu:
Hi @matt-FFFFFF ,

Thank you so much for taking your time to report this issue!!

I've opened a PR to fix this issue.

The correct parent_id format in this case should be `/providers/Microsoft.Management/managementGroups/{management group name}`
I suggest use `azapi` resource with `azurerm` resources together, like the following example:

```hcl
data "azurerm_subscription" "current" {
}

resource "azurerm_management_group" "example" {
  display_name = "myGroup"

  subscription_ids = [
[...]

PR #74, opened 2022-04-01
improve resource_id validation logics

ms-henglu:
To address issue:
https://github.com/Azure/terraform-provider-azapi/issues/73
Add a validation to prevent user setting `parent_id = ""`, which means "/" is the only allowed value for tenant scope top level resource.

To address issue:
https://github.com/Azure/terraform-provider-azapi/issues/72
I refactored the validation logics, into two parts.
1. Child resource type: verify if the parent_id's type matches the expected resource type, for example,
resource type is `Microsoft.Network/virtualNetworks/subnets`, we need to verify whether `parent_id` is an ID of a `Microsoft.Network/virtualNetworks`
2. Top level resource type: verify if `parent_id` provides a correct scope, like tenant scope `/`, subscription scope `/subscriptions/000-000-00`

### DIFF
--- a/internal/services/azapi_resource.go
+++ b/internal/services/azapi_resource.go
@@ -8,6 +8,7 @@ import (
 	"reflect"
 	"time"
 
+	"github.com/Azure/terraform-provider-azapi/internal/azure"
 	"github.com/Azure/terraform-provider-azapi/internal/azure/identity"
 	"github.com/Azure/terraform-provider-azapi/internal/azure/location"
 	"github.com/Azure/terraform-provider-azapi/internal/azure/tags"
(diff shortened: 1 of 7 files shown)


magodo (approved):
LGTM :+1: One minor suggestion is that we can consider leveraging the resource id parsing logic from azure sdk track2: https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/azcore/arm

ms-henglu:
> LGTM 👍 One minor suggestion is that we can consider leveraging the resource id parsing logic from azure sdk track2: https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/azcore/arm

Great idea! I created this to track that: https://github.com/Azure/terraform-provider-azapi/issues/75

LaurentLesle:
One of the limitation we have in azurerm is when we want to read let's say secrets from a keyvault outside of the scope of the parent_id it is failing saying the resource group does not exist. Cross subscription service chaining is something needed. This scenario is now very common in CAF and subscription democratisation and would suggest you don't implement too much logic and let the Azure control plane respond/fail based on RBAC or resource exist or not.

ms-henglu:
Hi @LaurentLesle , does it block some `Cross subscription service chaining` cases? If so, would you please file an issue and attach some configurations?

Issue #87, opened 2022-04-19
`use_msi` does not work


SudoSpartanDan:
The documentation states that using an MSI works with the provider. https://registry.terraform.io/providers/azure/azapi/latest/docs/guides/managed_service_identity#configuring-with-the-provider-block

But, in practice, it is not implemented in the provider configuration. https://github.com/Azure/terraform-provider-azapi/blob/159d8060e4fde94259200c6b1acffcc310fd4a7b/internal/provider/provider.go#L116-L129

Is this a feature we can expect to have? The azurerm provider supports this option.

ms-henglu:
Hi @SudoSpartanDan ,

Thank you so much for taking time to report this issue.

Yes, using an MSI works with azapi provider, and this feature: managed identity authentication is enabled in authentication chain by default, no need to configure `use_msi=true` like azurerm provider, I've raised a PR to update the document.

PR #89, opened 2022-04-20
remove use_msi from documents

ms-henglu:
To address https://github.com/Azure/terraform-provider-azapi/issues/87

Managed identity authentication is enabled in authentication chain by default, no need to expose property `use_msi`.

ref: https://github.com/Azure/terraform-provider-azapi/blob/main/vendor/github.com/Azure/azure-sdk-for-go/sdk/azidentity/default_azure_credential.go#L42

### DIFF
--- a/docs/guides/managed_service_identity.md
+++ b/docs/guides/managed_service_identity.md
@@ -60,9 +60,9 @@ Terraform can be configured to use managed identity for authentication in one of
 
 Using managed identity for Azure resources as authentication is enabled in `azapi` provider by default.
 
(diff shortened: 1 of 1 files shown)


magodo (changes requested):
Can we also remove the lines about `msi_endpoint` config as it is also not available for now: https://github.com/Azure/terraform-provider-azapi/pull/89/files#diff-3dabd9c8b8f5a6c14ee11c20b63ebf5ad3c5aa907cbf49bfb361a9438a34c5f0R65

ms-henglu:
@magodo , thanks! I've updated the doc, please take another look.

magodo (approved):

PR #92, opened 2022-04-25
suppress diff if zero value is configured in body and omitted in response

ms-henglu:
There'll be a plan-diff if user configured `some_property` to a zero value like empty string, empty array, empty map, and response omitted this property.

For example, in `Microsoft.ContainerService/managedClusters/agentPools@2022-01-02-preview`, if users want to reset `nodel_labels` to empty, users can use the following config:
```
nodeLabels = {}
```
And backend doesn't return `nodeLabels` in response, we should take this response as matched with configuration instead of diff.


magodo (approved):
LGTM 👍

PR #93, opened 2022-04-26
update golinter to v1.45.2

ms-henglu:


utils/json.go
```diff
@@ -214,7 +213,7 @@ func isZeroValue(value interface{}) bool {
 	case int, int32, int64, float32, float64:
 		return v == 0
 	case bool:
-		return v == false
+		return !v
```
magodo, line 216: Should we test whether `value` is `nil` at the entry of this function, in case user has specified `a = null` and the API doesn't return `a`.

ms-henglu: Good point! I'll add it.

magodo (approved):

PR #96, opened 2022-04-27
fix id format in imported resource

ms-henglu:
When import the `azapi_resource`, `api-version` is required as a query parameter, but this id should not be stored in state directly, `api-version` should be removed.


magodo (approved):
LGTM 👍

PR #104, opened 2022-05-12
improve error message for schema validation failure

ms-henglu:
The previous error message from schema validation failure is misleading, users might think it's a server-side error, so I added some messages suggesting how to fix it or bypass it.


magodo (approved):
LGTM :+1:

Issue #108, opened 2022-05-16
Provider doesn't honour ARM_CLIENT_ID for managed identity


Oechiih:
When using the AzAPI provider from an Azure VM with user assigned managed identity (for example a self-hosted Azure DevOps agent) it doesn't honour the ARM_CLIENT_ID environment variable to use a specific msi rather than a system-assigned identity. If no system-assigned msi is assigned the provider fails to get an access token. If there is a system-assigned msi the provider ignores the explicitly specified client_id and instead just uses the system-assigned one.
The regular azurerm provider supports user-assigned msi just fine.

Cheers
Jan

ms-henglu:
Hi @Oechiih ,

Thank you for taking time to report this issue.

It has been fixed in azidentity v1.0.0, I've raised a PR to update the dependency.

ms-henglu:
Hi @Oechiih ,

I just released the fix in https://github.com/Azure/terraform-provider-azapi/releases/tag/v0.2.1, would you please give it a try?

LaurentLesle:
I had a similar issue. version 0.2.1 fixed it. Thanks

ms-henglu:
Thanks @LaurentLesle , I'll close this issue.

PR #110, opened 2022-05-18
Update azcore and azidentity to v1.0.0

ms-henglu:
To address https://github.com/Azure/terraform-provider-azapi/issues/108


magodo (approved):
LGTM!

Issue #109, opened 2022-05-17
AKS - ManagedClusters not updating addonProfiles


gysiedebruyn:
I am attempting to disable the Kubernetes Dashboard in attempt to harding the AKS security and keep an eye on it using Azure Policy. The latest **azurerm** provider for AKS does not provide the functionality anymore so I thought I'd use the API directly using **azapi**. The result however is disappointing, it does not update the expected values of the AKS cluster.

This is how I call **azapi** in my Terraform (TF) code:
```
terraform {
    required_providers {
        azapi = {
            source = "azure/azapi"
            version = "=0.2.0"
        }
    }
}

resource "azapi_update_resource" "aks_dashboard_disable" {
    type = "Microsoft.ContainerService/managedClusters@2022-02-01"
    resource_id = azurerm_kubernetes_cluster.aks.id

    body = jsonencode({
        location = "australiaeast"
        properties =  {
            addonProfiles =  {
                kubeDashboard =  {
                    enabled = false
                    config =  null
                }
            }
        }
    })
}
```

The end result is no change in the addonProfiles property and there are no errors thrown and the status from TF is that has been successful.

When I attempt the exact same operation using Postman using the Azure REST API, it works as expected and updates _addonProfiles_ properties. The parameters and API versions are exactly the same for both methods. Here is the **curl** command exported from my Postman call:

```
curl --location --request PUT 'https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/jdb-test-rg/providers/Microsoft.ContainerService/managedClusters/jdb-test?api-version=2022-02-01' \
--header 'Content-Type: application/json' \
--data-raw '{
    "location": "australiaeast",
    "properties": {
        "addonProfiles": {
            "kubeDashboard": {
                "enabled": false,
                "config": null
            }
        }
    }
}'
```

As you can see, I use the same API version and parameters but get different results. I've been unable to gather any TF_LOG output as the amount of output "breaks" my console.

Anyone encountered this before or know why this happens? Or is it a genuine bug?

ms-henglu:
Hi @gysiedebruyn ,

Thank you for taking time to report this issue.

This is a bug caused by not handling additional properties properly and removed the config you updated by mistake. I'll fix it.
ref: https://github.com/Azure/terraform-provider-azapi/blob/main/internal/azure/types/object_type.go#L49

ms-henglu:
Hi @gysiedebruyn ,

[...]

PR #111, opened 2022-05-18
bugfix: additional property is the map's value type

ms-henglu:
To address https://github.com/Azure/terraform-provider-azapi/issues/109


magodo (approved):
LGTM! But maybe we shall add a test to capture this?

ms-henglu:
@magodo - Good advice! But there's already a testcase for it, and I've removed the hardcoding fix used to pass the case.

Issue #117, opened 2022-05-24
Need way of setting single run on child resources


adhodgson-wpp:
I have a situation where I want to create multiple child resources using a for_each block and its failing on the subsequent resources because the first request is blocking the rest of the requests from being actioned.

Here is some sample code:
```
resource "azapi_resource" "bgp" {
  for_each = var.bgp_peers
  depends_on = [
    azurerm_virtual_hub.hub
  ]
  type      = "Microsoft.Network/virtualHubs/bgpConnections@2021-05-01"
  name      = each.key
  parent_id = azurerm_virtual_hub.hub.id
  body = jsonencode({
    properties : {
      hubVirtualNetworkConnection : {
        id : azurerm_virtual_hub_connection.gateways.id
      },
      peerAsn : 65500,
      peerIp : each.value
    }
  })
}
```
In this case the first peer is created but I get an error when creating the rest of the peers:
```
│ Error: creating/updating "Resource: (ResourceId \"/subscriptions/[...]/resourceGroups/[...]/providers/Microsoft.Network/virtualHubs/[...]/bgpConnections/[...]\" / Api Version \"2021-05-01\")": GET https://management.azure.com/subscriptions/[...]/providers/Microsoft.Network/locations/uksouth/operations/[...]
│ --------------------------------------------------------------------------------
│ RESPONSE 200: 200 OK
│ ERROR CODE: ConflictError
│ --------------------------------------------------------------------------------
│ {
│   "status": "Failed",
│   "error": {
│     "code": "ConflictError",
│     "message": "The current operation could not be executed because it is already in progress.",
│     "details": []
│   }
│ }
│ --------------------------------------------------------------------------------
```

Looking at this a bit more and the code for the Azure provider I think we need some way of setting parallelism or a retry operation on the resource operation.

ms-henglu:
Hi @adhodgson-wpp ,

Thank you for taking time to report this issue, this is a good question.

A workaround for this is using option `-parallelism 1`, more info can be found [here](https://www.terraform.io/cli/commands/apply#parallelism-n)

adhodgson-wpp:
Hi. This option works and I should have mentioned it earlier. However I am running my Azapi resource as part of a larger project and don't want to set the serial operation for the rest of the resources that use the AzureRM provider. Thanks.

ms-henglu:
Hi @adhodgson-wpp ,

This feature has been released in v0.4.0, would you please give it a try? Thanks!

[...]

PR #123, opened 2022-05-31
azapi resource support `locks`

ms-henglu:
To address https://github.com/Azure/terraform-provider-azapi/issues/117

This property allows user to specify a list of ARM IDs, which is used to prevent creating/modifying resources at the same time.


magodo (approved):

docs/resources/azapi_update_resource.md
```diff
@@ -129,6 +129,8 @@ The following arguments are supported:
 }
 ```
 
+* `locks` - (Optional) A list of ARM resource IDs which are used to avoid create/modify azapi resources at the same time.
```
ms-henglu, line 132: Hi @grayzu ,

This is used to handle case like, attach multiple associations to a resource but the API only allows attaching one association at a time. Then user can set `locks` to the resource ID, so multiple associations won't be created at the same time.

 Do you think the property name and its description make sense?

magodo: Probably needs to mention it also affects the delete for the azapi_resource.md?

Issue #134, opened 2022-06-10
Add support to retrieve an object based on the resource id


LaurentLesle:
This request is to align the data.azapi_resource on a similar behaviour we already have with the az cli.

The use case is a public ssh key is in a remote subscription and I want to reference to create a linux virtual machine

```
# Get an sshkey from the resource id
az sshkey show --ids "/subscriptions/xxxxxxx/resourceGroups/yyyy/providers/Microsoft.Compute/sshPublicKeys/ssh-key"
```

Proposed implementation
```
provider "azapi" {
}

variable "ssh_public_key_id" {
  default = "/subscriptions/xxxxxxx/resourceGroups/yyyy/providers/Microsoft.Compute/sshPublicKeys/ssh-key"
}

data "azapi_resource" "ssh_public_key" {
  id    = var.ssh_public_key_id
  type  = "Microsoft.Compute/sshPublicKeys@2021-11-01"

  response_export_values = ["publicKey"]
}

output "ssh_public_key" {
   value = jsondecode(data.azapi_resource.ssh_public_key.output).publicKey
}
```

With this proposition, the name and parent_id are optional when id is set.

ms-henglu:
Hi @LaurentLesle ,

Thank you for opening this issue. I'll add this feature in next release.

ms-henglu:
Hi @LaurentLesle,

This feature has been released in v0.4.0, would you please give it a try? Thanks!

daveneeley:
`resource_id` won't validate for a resource like [Web Apps - List Host Keys Slot][1], which does not have a number of paths divisible by 2 (there are 13). Since name cannot be provided when resource_id is, there is no way to add this additional information.

[1]: https://docs.microsoft.com/en-us/rest/api/appservice/web-apps/list-host-keys-slot

PR #135, opened 2022-06-13
`azapi_resource` data source supports `resource_id`

ms-henglu:
To address: https://github.com/Azure/terraform-provider-azapi/issues/134

`azapi_resource` data source: Supports configuring `resource_id` as an alternative way to configure `name` and `parent_id`.

### DIFF
--- a/internal/services/azapi_resource_data_source.go
+++ b/internal/services/azapi_resource_data_source.go
@@ -26,15 +26,28 @@ func ResourceAzApiDataSource() *schema.Resource {
 
 		Schema: map[string]*schema.Schema{
 			"name": {
-				Type:         schema.TypeString,
-				Required:     true,
-				ValidateFunc: validation.StringIsNotEmpty,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ValidateFunc:  validation.StringIsNotEmpty,
+				RequiredWith:  []string{"parent_id"},
+				ConflictsWith: []string{"resource_id"},
+				AtLeastOneOf:  []string{"name", "resource_id"},
 			},
 
 			"parent_id": {
-				Type:     schema.TypeString,
-				Required: true,
-				//ValidateFunc: validate.AzureResourceID,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ValidateFunc:  validation.StringIsNotEmpty,
+				RequiredWith:  []string{"name"},
+				ConflictsWith: []string{"resource_id"},
+			},
+
+			"resource_id": {
+				Type:          schema.TypeString,
+				Optional:      true,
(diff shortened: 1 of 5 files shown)


internal/services/azapi_resource_data_source.go
```diff
@@ -26,15 +26,28 @@ func ResourceAzApiDataSource() *schema.Resource {
...
 			"name": {
-				Type:         schema.TypeString,
-				Required:     true,
-				ValidateFunc: validation.StringIsNotEmpty,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ValidateFunc:  validation.StringIsNotEmpty,
+				RequiredWith:  []string{"parent_id"},
+				ConflictsWith: []string{"resource_id"},
+				AtLeastOneOf:  []string{"name", "resource_id"},
```
magodo, line 34: This can be merged into: `ExactlyOneOf`?

internal/services/azapi_resource_data_source.go
```diff
@@ -26,15 +26,28 @@ func ResourceAzApiDataSource() *schema.Resource {
...
 
 			"parent_id": {
-				Type:     schema.TypeString,
-				Required: true,
-				//ValidateFunc: validate.AzureResourceID,
+				Type:          schema.TypeString,
+				Optional:      true,
+				ValidateFunc:  validation.StringIsNotEmpty,
+				RequiredWith:  []string{"name"},
+				ConflictsWith: []string{"resource_id"},
```
magodo, line 42: Then these might be redundent?

internal/services/azapi_resource_data_source.go
```diff
@@ -26,15 +26,28 @@ func ResourceAzApiDataSource() *schema.Resource {
...
+			},
+
+			"resource_id": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ValidateFunc:  validate.AzureResourceID,
+				ConflictsWith: []string{"name", "parent_id"},
+				AtLeastOneOf:  []string{"name", "resource_id"},
```
magodo, line 50: This can be merged into: `ExactlyOneOf`?

magodo (changes requested):
Thanks for this PR!
I have left some additional mostly minor comments that once addressed this should be good to merge 👍

internal/services/azapi_resource_data_source.go
ms-henglu: Good catch, I'll update it.

magodo: The ConfclitsWith can also be removed than.

ms-henglu: Yes, and it has been removed.

magodo (approved):
LGTM!

PR #146, opened 2022-07-05
Add `append_existing_values` capability

### LINKED ISSUE #145: Feature request: Allow appending properties from existing resource
Hi,

In the use case of deploying virtual networks using AzAPI, but not deploying subnets. Any existing subnet resources are deleted when updating the virtual network, e.g. updating the tags.

This is not specifically an AzAPI problem, it is more the behaviour of the Azure network resource provider.

```terraform
resource "azapi_resource" "vnet" {
  parent_id = "/subscriptions/00000000-0000-0000-000000000000/resourceGroups/azapitest"
  type = "Microsoft.Network/virtualNetworks@2021-08-01"
  name = "vnet"
  location = "northeurope"
    body = jsonencode({
    properties = {
      addressSpace = {
        addressPrefixes = [
          "10.0.0.0/16"
        ]
      }
    }
  })
  tags = {
    deployedBy = "azapitest2"
  }
}
```

What I would like to do is propose a feature that takes a JSON path from the existing resource, if it exists, then appends it to the body of the PUT request.

Could be implemented in a similar way to `response_export_values`, in that a list of JSON keys can be provided for appending to the PUT request.

The order of events would be:

- If resource exists...
- Perform a GET
- for each of the supplied JSON paths, extract the values from the GET request body
- Append said values to the prepared PUT body, if they do not already exist.

I see that much of the JSON path logic is already present in `json.go`. I would be prepared to give this a go with a PR if you think it's a good idea?

matt-FFFFFF:
Fixes #145

Attn: @ms-henglu would you consider something like this?

Adds capability to append existing ARM values to the PUT request.

This is useful when dealing with Resource Providers that have odd behaviours.

### DIFF
--- a/docs/resources/azapi_resource.md
+++ b/docs/resources/azapi_resource.md
@@ -81,8 +81,8 @@ output "quarantine_policy" {
 ## Arguments Reference
 
 The following arguments are supported:
-* `name` - (Required) Specifies the name of the azure resource. Changing this forces a new resource to be created. 
-* `parent_id` - (Required) The ID of the azure resource in which this resource is created. Changing this forces a new resource to be created. It supports different kinds of deployment scope for **top level** resources: 
+* `name` - (Required) Specifies the name of the azure resource. Changing this forces a new resource to be created.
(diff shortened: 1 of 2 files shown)


matt-FFFFFF:
Closing as have workaround using azapi_update_resource

PR #148, opened 2022-07-07
Client initialization option set cloud config

magodo:
Client initialization option set cloud config to support clouds othre than public.


ms-henglu (approved):
LGTM!

Issue #57, opened 2022-03-01
New resource `azapi_resource_action` - manage any azure resource action


ms-henglu:
This resource provides more flexibility, allows user to manage operations of an ARM resource, for example: list properties, or restart/stop a VM.

## v2 design
In the latest design, one `azapi_operation` can only manage one operation, this is simpler than previous design, and user can still toggle different status by using multiple operations with `count=0/1`. 

And there're other scenarios that this operation resource can help.
In short, azapi_operation enhanced azapi_resource/update_resource, user can add customized way to update azapi_resource, or a customized source for azapi_update_resource.
 
Example 1. The maria db only supports PATCH method to update some of its properties, so we use azapi_operation to manage the properties that needs to be updated by PATCH method.
This enables user to update resources which don’t support PUT update.
 
```
resource "azapi_resource" "server" {
  type      = "Microsoft.DBforMariaDB/servers@2018-06-01-preview"
  parent_id = azurerm_resource_group.test.id
  name      = "myServer"
  location  = "westus"
  body = jsonencode({
    properties = {
      administratorLogin = "adminuser"
      administratorLoginPassword = "<administratorLoginPassword>"
      createMode = "Default"
    }
  })
}
 
resource "azapi_operation" "server_configure" {
  type        = "Microsoft.DBforMariaDB/servers@2018-06-01-preview"
  resource_id = azapi_resource.server.id
  operation   = ""
  method      = "PATCH"
  body = jsonencode({
    properties = {
      administratorLoginPassword = "<administratorLoginPassword>"
      sslEnforcement             = "Disabled"
    }
  })
}
```
 
Example 2. This is used to address case like [azapi_update_resource - Put vs Patch · Issue #97 · Azure/terraform-provider-azapi (github.com)](https://github.com/Azure/terraform-provider-azapi/issues/97)
The problem is when user updates appsettings, appsettings can not be retrieved by GET because it contains credentials, it can only be retrieved by a POST request.
So, azapi_update_resource couldn’t work properly, because it can’t get the correct existing properties. But with azapi_operation, user can add a customized operation to get existing properties, and use hcl merge function to combine the existing properties and newly added ones.

```
resource "azapi_operation" "list_config" {
  type                   = "Microsoft.Web/sites/config@2021-02-01"
  resource_id            = "${azurerm_function_app.fnapp.id}/config/appsettings"
  operation              = "list"
  method                 = "POST"
  response_export_values = ["*"]
}
 

[...]

PR #150, opened 2022-07-12
new resource `azapi_opration`

ms-henglu:
To address: https://github.com/Azure/terraform-provider-azapi/issues/57

### DIFF
--- a/docs/data-sources/azapi_operation.md
+++ b/docs/data-sources/azapi_operation.md
@@ -1,0 +1,110 @@
+---
+subcategory: ""
+layout: "azapi"
+page_title: "Generic Azure Resource Operation Data Source: azapi_operation"
+description: |-
+  Perform resource function to get information from an existing azure resource
+---
+
+# azapi_operation
+
+This resource can perform resource function to get information form any existing Azure resource manager resource.
+
+## Example Usage
+
+```hcl
+terraform {
+  required_providers {
+    azapi = {
+      source = "Azure/azapi"
+    }
+  }
+}
+
+provider "azapi" {
+}
+
+provider "azurerm" {
+  features {}
+}
+
+resource "azurerm_resource_group" "example" {
+  name     = "example-rg"
+  location = "west europe"
+}
+
+resource "azurerm_automation_account" "example" {
+  name                = "example-account"
+  resource_group_name = azurerm_resource_group.example.name
+  location            = azurerm_resource_group.example.location
+  sku_name            = "Basic"
+}
+
+data "azapi_operation" "example" {
+  type                   = "Microsoft.Automation/automationAccounts@2021-06-22"
+  resource_id            = azurerm_automation_account.example.id
+  operation              = "listKeys"
+  response_export_values = ["*"]
+}
+```
+
+## Arguments Reference
+
+The following arguments are supported:
+
+* `type` - (Required) It is in a format like `<resource-type>@<api-version>`. `<resource-type>` is the Azure resource type, for example, `Microsoft.Storage/storageAccounts`.
+  `<api-version>` is version of the API used to manage this azure resource.
+
+* `resource_id` - (Required) The ID of an existing azure source.
+
+* `operation` - (Optional) The name of the resource operation. It's also possible to make Http requests towards the resource ID if leave this field empty.
+
+---
+* `body` - (Optional) A JSON object that contains the request body.
+
+* `method` - (Optional) Specifies the Http method of the azure resource operation. Allowed values are `POST`, `PATCH`, `GET`, `PUT`, `DELETE`, `CONNECT`, `HEAD`, `OPTIONS` and `TRACE`. Defaults to `POST`.
+
+* `response_export_values` - (Optional) A list of path that needs to be exported from response body.
+  Setting it to `["*"]` will export the full response body.
+  Here's an example. If it sets to `["keys"]`, it will set the following json to computed property `output`.
+```
+{
+  "keys": [
+    {
+      "KeyName": "Primary",
+      "Permissions": "Full",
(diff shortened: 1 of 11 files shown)


docs/data-sources/azapi_operation.md
```diff
@@ -0,0 +1,110 @@
...
+  `<api-version>` is version of the API used to manage this azure resource.
+
+* `resource_id` - (Required) The ID of an existing azure source.
+
+* `operation` - (Optional) The name of the resource operation. It's also possible to make Http requests towards the resource ID if leave this field empty.
+
+---
+* `body` - (Optional) A JSON object that contains the request body.
+
+* `method` - (Optional) Specifies the Http method of the azure resource operation. Allowed values are `POST`, `PATCH`, `GET`, `PUT`, `DELETE`, `CONNECT`, `HEAD`, `OPTIONS` and `TRACE`. Defaults to `POST`.
```
magodo, line 65: Should we narrow down the allowed enums for this `method`, as it is only used for query?

internal/clients/resource_client.go
```diff
@@ -173,3 +174,56 @@ func (client *ResourceClient) deleteCreateRequest(ctx context.Context, resourceI
...
+		return nil, err
+	}
+	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusCreated, http.StatusAccepted) {
+		return nil, runtime.NewResponseError(resp)
+	}
+	return resp, nil
+}
+
+func (client *ResourceClient) actionCreateRequest(ctx context.Context, resourceID string, action string, apiVersion string, method string, body interface{}) (*policy.Request, error) {
+	urlPath := fmt.Sprintf("/%s", resourceID)
```
magodo, line 213: Is `resourceID` already started with a slash?

internal/clients/resource_client.go
```diff
@@ -173,3 +174,56 @@ func (client *ResourceClient) deleteCreateRequest(ctx context.Context, resourceI
...
+	if len(action) != 0 {
+		urlPath = fmt.Sprintf("/%s/%s", resourceID, action)
```
magodo, line 215: Same here.

internal/clients/resource_client.go
```diff
@@ -173,3 +174,56 @@ func (client *ResourceClient) deleteCreateRequest(ctx context.Context, resourceI
...
+	}
+	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusCreated, http.StatusAccepted) {
+		return nil, runtime.NewResponseError(resp)
+	}
+	return resp, nil
+}
+
+func (client *ResourceClient) actionCreateRequest(ctx context.Context, resourceID string, action string, apiVersion string, method string, body interface{}) (*policy.Request, error) {
+	urlPath := fmt.Sprintf("/%s", resourceID)
+	if len(action) != 0 {
```
magodo, line 214: I see you use `len()` to check emptyness of a string at a couple of places, why not just `action != ""`?

internal/services/azapi_operation_data_source.go
```diff
@@ -0,0 +1,109 @@
...
+
+			"operation": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"method": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "POST",
```
magodo, line 51: There is no validation list defined, but the document said there is..

internal/services/azapi_operation_resource.go
```diff
@@ -0,0 +1,148 @@
...
+		},
+
+		CustomizeDiff: func(ctx context.Context, d *schema.ResourceDiff, meta interface{}) error {
+			if d.HasChange("response_export_values") {
+				d.SetNewComputed("output")
+			}
+			old, new := d.GetChange("body")
+			if utils.NormalizeJson(old) != utils.NormalizeJson(new) {
+				d.SetNewComputed("output")
+			}
```
magodo, line 103: This can be removed if we define the StateFunc for the `body`?

docs/resources/azapi_operation.md
```diff
@@ -0,0 +1,133 @@
...
+
+* `resource_id` - (Required) The ID of an existing azure source. 
+
+* `operation` - (Optional) The name of the resource operation. It's also possible to make Http requests towards the resource ID if leave this field empty.
+
+---
+
+* `body` - (Optional) A JSON object that contains the request body.
+
+* `method` - (Optional) Specifies the Http method of the azure resource operation. Allowed values are `POST`, `PATCH`, `GET`, `PUT`, `DELETE`, `CONNECT`, `HEAD`, `OPTIONS` and `TRACE`. Defaults to `POST`.
```
magodo, line 85: At least the `GET` is not a fit in here.

docs/resources/azapi_operation.md
```diff
@@ -0,0 +1,133 @@
...
+subcategory: ""
+layout: "azapi"
+page_title: "Generic Azure Resource Operation: azapi_operation"
+description: |-
+  Perform resource function of an existing azure resource
+---
+
+# azapi_operation
+
+This resource can manage any Azure resource manager resource operation.
```
magodo, line 11: Can we emphasize when to use the resource and when to use the data source? From my understanding, the operation that is only used for reading should use the data source.

Also, we shall also add more words in the data source document.

magodo (changes requested):
Thanks for this PR!
I've taken a look through and left some comments inline, but this is mostly looking good to me 👍

ms-henglu:
Thanks @magodo for code review, I've addressed these comments, please take another look, thanks!

magodo (approved):

magodo:
@ms-henglu LGTM!

Issue #144, opened 2022-07-04
DefaultAzureCredential authentication failed


HametAk:
I have a big issue with a module which uses AzApi.

I get the following error:
It is not possible for me to do a terraform apply with AzApi.

I already tried to set environment variables manually and only do az login on specific tenant.
This is reproducible in my company.

The environment used is the Azure Virtual Desktop and error only occurs there.
![MicrosoftTeams-image](https://user-images.githubusercontent.com/70479098/177104305-468a9238-4cb1-4fbc-b752-07ad9c3d5cba.png)

ms-henglu:
Hi @HametAk ,

It seems that you didn't configure the credentials properly, please check https://registry.terraform.io/providers/Azure/azapi/latest/docs/guides/serivce_principal_client_sercret and see if there're some fields are missed?

HametAk:
Hi @ms-henglu ,
thanks for quick reply.

Is this necessary? Because I am not using a service principal or is it mandatory/recommended to use one?

ms-henglu:
Hi @HametAk - No, you can check the [guides](https://registry.terraform.io/providers/Azure/azapi/latest/docs) on the left side, there're other ways to authenticate with azure.

HametAk:
I am just using simple "az login -t <tenant>" on my other environment which works just fine.
It's just an issue when using the Azure Cloud Desktop.
I can do every other terraform task except AzAPI

ms-henglu:
@HametAk

> I can do every other terraform task except AzAPI

You mean, `azurerm` provider works fine?

HametAk:
@ms-henglu 
Yes it works fine.

HametAk:
Following Code was used:
```

terraform {
  required_providers {
    azapi = {
      source  = "Azure/azapi"
      version = "~> 0.3.0"
    }
  }
}

resource "azapi_update_resource" "subscription_tags" {
  type      = "Microsoft.Resources/tags@2021-04-01"
  parent_id = "/subscriptions/${var.subscription_id}"
  name      = "default"
  body = jsonencode({
    properties = {
      tags = var.tags
    }
  })
}
```

andrewPoland:
I ran into this issue too when running from an Azure hosted VM. The VM has no System or User Assigned Identities and I'm attempting to use `az login` for authentication. 

Reverting back to version `0.2.0` allowed me to create the resources using credentials from `az login`.  

[...]

PR #151, opened 2022-07-12
bugfix: do not set env if values are empty

ms-henglu:
To address: https://github.com/Azure/terraform-provider-azapi/issues/144

Root cause: If an invalid `AZURE_CLIENT_ID` is set, managed identity authentication will fail as its configuration has error, which is fatal, then the azure cli authentication is skipped.

refs: https://github.com/Azure/azure-sdk-for-go/issues/18352#issuecomment-1160987065


magodo (approved):
LGTM! Though it is more like a bug in the azcore.

PR #166, opened 2022-08-15
rename `azapi_operation` to `azapi_resource_action`

ms-henglu:
related to https://github.com/Azure/terraform-provider-azapi/issues/57


magodo:
I'm wondering the reason for the renaming, as there are operations that are supported but doesn't belong to any resource?

ms-henglu:
> I'm wondering the reason for the renaming, as there are operations that are supported but doesn't belong to any resource?

Good question, but most of them belong to one resource, and based on the arm wiki, they refer these operations as actions.

magodo (approved):

PR #168, opened 2022-08-16
action resource's id should be the request path

ms-henglu:
The IDs of `azapi_resource_action` resource and data source were set to the IDs of the targeting resources, but it's better to use request path(combining action).

### DIFF
--- a/internal/services/azapi_resource_action_data_source.go
+++ b/internal/services/azapi_resource_action_data_source.go
@@ -109,7 +109,11 @@ func resourceResourceActionDataSourceRead(d *schema.ResourceData, meta interface
 		return fmt.Errorf("performing action %s of %q: %+v", actionName, id, err)
 	}
 
-	d.SetId(id.ID())
+	resourceId := id.ID()
+	if actionName != "" {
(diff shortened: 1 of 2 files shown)


internal/services/azapi_resource_action_data_source.go
```diff
@@ -109,7 +109,11 @@ func resourceResourceActionDataSourceRead(d *schema.ResourceData, meta interface
 		return fmt.Errorf("performing action %s of %q: %+v", actionName, id, err)
 	}
 
-	d.SetId(id.ID())
+	resourceId := id.ID()
+	if actionName != "" {
+		resourceId = fmt.Sprintf("%s/%s", id.ID(), resourceId)
```
magodo, line 114: Should this be?
```suggestion
		resourceId = fmt.Sprintf("%s/%s", id.ID(), actionName)
```

magodo (changes requested):

internal/services/azapi_resource_action_data_source.go
ms-henglu: Good catch! I'll fix it.

magodo (approved):

Issue #178, opened 2022-09-02
feature request: support locks for r/azapi_resource_action


t3mi:
This is a feature request to add locks support for the `azapi_resource_action` resource so that multiple actions could be handled successfully against single resource.

ms-henglu:
Hi @t3mi ,

Thank you for taking time to report this issue!

I'll add this feature in the next release. Would you please share me more details about your case? Thanks!

t3mi:
I'm trying to automatically approve private endpoint connection but its creation is done in the Azure backend and sometimes results in creation of only one endpoint for multiple "creation" calls. So current behavior without the locks results in the error for second approval command:
```hcl
Operation PutPrivateEndpointConnectionOperation (<GUID>) was canceled and superseded by operation PutPrivateEndpointConnectionOperation (<GUID>).
```

PR #180, opened 2022-09-05
`azapi_resource_action` - `locks`

ms-henglu:
Fixes https://github.com/Azure/terraform-provider-azapi/issues/178

### DIFF
--- a/docs/resources/azapi_resource_action.md
+++ b/docs/resources/azapi_resource_action.md
@@ -84,6 +84,8 @@ The following arguments are supported:
 
 * `body` - (Optional) A JSON object that contains the request body.
 
+* `locks` - (Optional) A list of ARM resource IDs which are used to avoid modify azapi resources at the same time.
+
 * `method` - (Optional) Specifies the Http method of the azure resource action. Allowed values are `POST`, `PATCH`, `PUT` and `DELETE`. Defaults to `POST`.
 
 * `response_export_values` - (Optional) A list of path that needs to be exported from response body.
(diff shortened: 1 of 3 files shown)


docs/resources/azapi_resource_action.md
```diff
@@ -84,6 +84,8 @@ The following arguments are supported:
 
 * `body` - (Optional) A JSON object that contains the request body.
 
+* `locks` - (Optional) A list of ARM resource IDs which are used to avoid modify azapi resources at the same time.
```
magodo, line 87: This looks like not necessarily need to be a resource ID?

internal/services/azapi_resource_action_resource.go
```diff
@@ -121,6 +131,12 @@ func resourceResourceActionCreateUpdate(d *schema.ResourceData, meta interface{}
 	}
 
 	log.Printf("[INFO] request body: %v\n", body)
+
+	for _, id := range d.Get("locks").([]interface{}) {
```
magodo, line 135: Should we also support it for delete?

magodo (changes requested):

docs/resources/azapi_resource_action.md
ms-henglu: Yes, but I want to align this description with `azapi_resource` and users might get confused if it says any strings are acceptable.

internal/services/azapi_resource_action_resource.go
magodo: Oh, missed it was for action...

magodo (approved):

Issue #101, opened 2022-05-06
Support for OIDC Authentication


ms-henglu:
Use OpenID Connect within your workflows to authenticate with Azure.

Blocked by https://github.com/Azure/azure-sdk-for-go/issues/15615

doc: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-azure
similar issue: https://github.com/hashicorp/terraform-provider-azurerm/issues/16554

ms-henglu:
https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/azidentity doesn't support OIDC yet.

Roadmap: https://github.com/Azure/azure-sdk-for-go/issues/16728

petr-stupka:
Helper already available via https://github.com/hashicorp/go-azure-helpers/pull/115

Terraform 1.2.0 introduced support for backend authentication OIDC as well

LaurentLesle:
@ms-henglu can you check if you can progress with this issue? Looks like the dependency has been closed https://github.com/Azure/azure-sdk-for-go/issues/16728

jamescrowley:
Just got blocked by this as we're using OIDC for GitHub Actions on a new project. We're setting the `ARM_USE_OIDC` environment variable, by the way, so if you can support that in addition to any properties on the provider itself that would be fab. Many thanks.

ms-henglu:
@LaurentLesle - Thanks! I just checked again, it seems they're tracking this feature by https://github.com/Azure/azure-sdk-for-go/issues/15615 according to https://github.com/Azure/azure-sdk-for-go/issues/16728#issuecomment-1119759637.

thomas-vercoutre-nzta:
Any news on that feature ?

dschniepp:
Does it make sense to provide a temporarily solution based on the upcoming release, with the workaround described here https://github.com/Azure/azure-sdk-for-go/issues/15615#issuecomment-1211012677? @ms-henglu would you be interested in this as a contribution?

ms-henglu:
Hi @dschniepp ,

Thanks for the suggestion! I'll give it a try when the azidentity v1.2.0 stable version released in the next week.

dschniepp:
Thanks, @ms-henglu if you need support feel free to loop me in.

AdamCoulterOz:
@grayzu was this planned for `1.1`? looks like it didn't make it?

ms-henglu:
Hi all,

[...]

PR #239, opened 2023-01-05
add OIDC authentication support

adelinn:
I added support for the following configuration parameters:
`use_oidc` - Whether to allow OpenID Connect to be used for authentication. Defaults to false.
`oidc_token_file_path` - The path to the file where the token from the OIDC provider is. This is useful for use with Kubernetes.
`oidc_token` - The token from the OIDC provider.
`oidc_request_url` - The URL for the OIDC provider from which to request an ID token.
`oidc_request_token` - The bearer token for the request to the OIDC provider.

The parameters work in the same way they do in the azurerm provider, the only difference would be that the order of the credentials that are tried is changed. The OidcCredential, which is the one added in this pull request, is checked first, then the default credential is checked, which actually checks the other credentials in the already documented order.

I didn't update the documentation files. I leave that to someone else.
Also the code changes that I made seem not to be in the correct location if I think about best practices, but it works...

This pull request would solve https://github.com/Azure/terraform-provider-azapi/issues/101

PS: This is one of my first pull requests to a public project and I'm a beginner in go lang.


ms-henglu:
Hi @adelinn ,

Thank you so much for opening this PR and apologies for the delayed review.

My teammate was working on this feature and now it's supported the OIDC authentication in https://github.com/Azure/terraform-provider-azapi/pull/240. 

Since we're unable to accept this contribution, whilst I'd like to thank you for this PR I'm going to close this for the moment.

Thanks!

PR #250, opened 2023-01-29
`azapi_resource` - ignore secrets which are returned in asterisks

ms-henglu:
By Azure rest API guideline, sensitive properties should be omitted in GET response, but some of them are returned in asterisks. This PR will ignore those secrets in asterisks.

### DIFF
--- a/utils/json.go
+++ b/utils/json.go
@@ -3,6 +3,7 @@ package utils
 import (
 	"encoding/json"
 	"fmt"
+	"regexp"
 	"strings"
 )
 
@@ -92,6 +93,9 @@ func GetUpdatedJson(old interface{}, new interface{}, option UpdateJsonOption) i
 	case string:
 		if newStr, ok := new.(string); ok {
 			if option.IgnoreCasing && strings.EqualFold(oldValue, newStr) {
+				return oldValue
+			}
+			if option.IgnoreMissingProperty && regexp.MustCompile(`^\**$`).MatchString(newStr) {
 				return oldValue
 			}
 		}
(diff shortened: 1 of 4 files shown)


utils/json.go
```diff
@@ -94,6 +95,9 @@ func GetUpdatedJson(old interface{}, new interface{}, option UpdateJsonOption) i
 			if option.IgnoreCasing && strings.EqualFold(oldValue, newStr) {
 				return oldValue
 			}
+			if option.IgnoreMissingProperty && regexp.MustCompile(`^\**$`).MatchString(newStr) {
```
magodo, line 98: What if there is an attribute that is set as `***` as input? Also, should the regexp be `^\*+$`, as otherwise empty string will also be matched?

magodo (commented):
Shall we make the `ignore_missing_property` something similar to [this](https://registry.terraform.io/providers/magodo/restful/latest/docs/resources/resource#write_only_attrs) to accept a list of attribute path, so that users can explicitly mark which attributes to ignore in GET, to avoid ignoring attribues by mistakes?

utils/json.go
ms-henglu: > What if there is an attribute that is set as *** as input?
This should not be a common case, user sets "***" but it returns different amount of *.

> should the regexp be ^\*+$, as otherwise empty string will also be matched?
Yes, I've updated it.

magodo (approved):

ms-henglu:
> Shall we make the ignore_missing_property something similar to [this](https://registry.terraform.io/providers/magodo/restful/latest/docs/resources/resource#write_only_attrs) to accept a list of attribute path, so that users can explicitly mark which attributes to ignore in GET, to avoid ignoring attribues by mistakes?

Yes, we might want to introduce this feature if there's an issue complaining about it.

PR #257, opened 2023-02-09
support `partial mode` to avoid writing diff to state while `Update` error

teowa:
Update with wrong API version will write wrong value to state file, which leads to state refresh error in later apply.

related to https://github.com/hashicorp/terraform-plugin-sdk/issues/476
use the `partial mode`, see [doc](https://www.hashicorp.com/blog/writing-custom-terraform-providers):
> If the Update callback returns with or without an error, the full state is saved. If the ID becomes blank, the resource is destroyed (even within an update, though this shouldn't happen except in error scenarios).

> Partial mode is a mode that can be enabled by a callback that tells Terraform that it is possible for partial state to occur. When this mode is enabled, the provider must explicitly tell Terraform what is safe to persist and what is not.

```
TF_ACC=1 go test -v ./internal/services -parallel 20 -test.run=TestAccGenericResource_invalidVersionUpdate -timeout 1440m -ldflags="-X=github.com/Azure/terraform-provider-azapi/version.ProviderVersion=acc"
=== RUN   TestAccGenericResource_invalidVersionUpdate
=== PAUSE TestAccGenericResource_invalidVersionUpdate
=== CONT  TestAccGenericResource_invalidVersionUpdate
--- PASS: TestAccGenericResource_invalidVersionUpdate (99.38s)
PASS
ok      github.com/Azure/terraform-provider-azapi/internal/services     101.030s

```

### DIFF
--- a/internal/services/azapi_resource.go
+++ b/internal/services/azapi_resource.go
@@ -281,6 +281,10 @@ func resourceAzApiResourceCreateUpdate(d *schema.ResourceData, meta interface{})
 	ctx, cancel := tf.ForCreateUpdate(meta.(*clients.Client).StopContext, d)
 	defer cancel()
 
+	if !d.IsNewResource() {
+		d.Partial(true)
+	}
+
 	config := d.GetRawConfig()
 	var resourceName string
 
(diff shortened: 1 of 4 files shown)


teowa:

ms-henglu (changes requested):
Hi @teowa ,

Thank you so much for making this PR! It looks good to me, but would you please also make this change to other azapi resources? Thanks!

ms-henglu (approved):
LGTM！

PR #267, opened 2023-03-14
Fix ADF Managed Airflow template

bramvdklinkenberg:
With this PR I fix the Managed Airflow IntegrationRuntime template. 

- The airflow version was not correctly specified
- I added the other airflow properties
- I updated the version to the latest version


ms-henglu (approved):
Thanks @bramvdklinkenberg for making this! LGTM.

Issue #316, opened 2023-07-17
Support provider level actions in `azapi_resource_action`


ms-henglu:
There're some provider level actions and they couldn't be supported by `azapi_resource_action` very well. Because we expect the `resource_id` is an azure resource id, but the provider level actions don't have a resource id and there's proper type for them.

Here's a proposal to support provider level actions in `azapi_resource_action`:

For example, `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Foo/queries`, the `resource_id` should be `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Foo`, and the `type` should be `Microsoft.Foo@api-version`.

The caveat is that the `resource_id` is not a real resource id, and it needs to be built by the user like `"${azurerm_resource_group.example.id}/providers/Microsoft.Foo"`.

ms-henglu:
related issue: https://github.com/Azure/terraform-provider-azapi/issues/311

ms-henglu:
Close this issue since it's released in v1.8.0

Issue #299, opened 2023-06-08
Feature request: `azapi_resource_action` not flexible enough to register resource providers or get provider data


matt-FFFFFF:
## Scenario:

Customer wants to pre-register resource providers using azapi.

Customer also wants to get provider data using provider api.

I realise that azapi will try to do this for any resource that it tries to deploy https://github.com/Azure/terraform-provider-azapi/issues/82#issuecomment-1095709598

However this does not cover the customer scenario where they want to pre-register providers for another team to use (perhaps not with TF). OR when a customer wants to get provider data to determine availability zone support for resources.

## Attempted configuration:

```terraform
resource "azapi_resource_action" "provreg" {
  resource_id = "/subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.Compute"
  type        = "Microsoft.Resources/providers@2021-04-01"
  method      = "POST"
  action      = "register"
}
```

I tried this but the HTTP Client within the provider changed the URL to this. You see it inserted the supplied `type` into the request:

```
│ --------------------------------------------------------------------------------
│ RESPONSE 404: 404 Not Found
│ ERROR CODE UNAVAILABLE
│ --------------------------------------------------------------------------------
│ {
│   "message": "No HTTP resource was found that matches the request URI 'https://management.azure.com/subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.Resources/providers/Microsoft.Compute/register?api-version=2021-04-01'."
│ }
│ --------------------------------------------------------------------------------
```

## Solution requested:

I would like more control over the REST request to enable a request to a URI of my choosing, in this case: `/subscriptions/{subscriptionId}/providers/{resourceProviderNamespace}/register?api-version=2021-04-01`.

AND 

GET `/subscriptions/{subscriptionId}/providers/{resourceProviderNamespace}?api-version=2021-04-01`

It would be good if there was an additional attribute for both data and managed resource `azapi_resource_action`... like this:

### Managed resource

```terraform
resource "azapi_resource_action" "provreg" {
  uri         = "/subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.Compute/register"
  api_version = "2021-04-01"
  method      = "POST"
  body        = jsonencode({})
}
```

### Data source

```terraform
# This gets the output from the compute provider
# This us useful in allowing us to find if a resource supports
# availability zones or not
data "azapi_resource_action" "compute_provider" {
[...]

PR #317, opened 2023-07-20
`azapi_resource_action` - supports provider level actions

ms-henglu:
related issues: https://github.com/Azure/terraform-provider-azapi/issues/299, https://github.com/Azure/terraform-provider-azapi/issues/316

This PR does the following things:
1. The `type` supports the resource provider name as the resource type for the provider IDs, for example, `Microsoft.Network`, to enable provider level actions with `azapi_resource_action` resource/data source.
2. Fix the bug that resource IDs for type  `Microsoft.Resources/providers` for provider IDs as not parsed correctly. It's used to register/unregister RP or fetch information of RPs.

```

=== RUN   TestAccActionResource_basic
=== PAUSE TestAccActionResource_basic
=== CONT  TestAccActionResource_basic
--- PASS: TestAccActionResource_basic (156.84s)
=== RUN   TestAccActionResource_registerResourceProvider
=== PAUSE TestAccActionResource_registerResourceProvider
=== CONT  TestAccActionResource_registerResourceProvider
--- PASS: TestAccActionResource_registerResourceProvider (35.90s)
=== RUN   TestAccActionResource_providerAction
=== PAUSE TestAccActionResource_providerAction
=== CONT  TestAccActionResource_providerAction
--- PASS: TestAccActionResource_providerAction (45.06s)
PASS


=== RUN   TestAccActionDataSource_basic
=== PAUSE TestAccActionDataSource_basic
=== CONT  TestAccActionDataSource_basic
=== RUN   TestAccActionDataSource_providerPermissions
=== PAUSE TestAccActionDataSource_providerPermissions
=== CONT  TestAccActionDataSource_providerPermissions
--- PASS: TestAccActionDataSource_providerPermissions (46.08s)
=== RUN   TestAccActionDataSource_providerAction
=== PAUSE TestAccActionDataSource_providerAction
=== CONT  TestAccActionDataSource_providerAction
--- PASS: TestAccActionDataSource_providerAction (35.63s)
--- PASS: TestAccActionDataSource_basic (150.61s)
PASS


```

### DIFF
--- a/docs/data-sources/azapi_resource_action.md
+++ b/docs/data-sources/azapi_resource_action.md
@@ -50,6 +50,25 @@ data "azapi_resource_action" "example" {
 }
 ```
 
+Here's an example to use `azapi_resource_action` resource to perform a provider action.
+
+```hcl
+terraform {
+  required_providers {
+    azapi = {
+      source = "Azure/azapi"
+    }
+  }
+}
+
+resource "azapi_resource_action" "test" {
+  type        = "Microsoft.Resources/providers@2021-04-01"
+  resource_id = "/subscriptions/{subscription_id}/providers/Microsoft.Compute"
+  action      = "register"
+  method      = "POST"
+}
+```
+
 ## Arguments Reference
 
 The following arguments are supported:
--- a/docs/data-sources/azapi_resource_action.md
+++ b/docs/data-sources/azapi_resource_action.md
@@ -67,6 +67,16 @@ resource "azapi_resource_action" "test" {
   action      = "register"
   method      = "POST"
 }
+
+resource "azapi_resource_action" "test" {
+  type        = "Microsoft.Databox@2022-10-01"
+  resource_id = "/subscriptions/{subscription_id}/providers/Microsoft.Databox"
+  action      = "validateAddress"
+  method      = "POST"
+  body = jsonencode({
+    // ...
+  })
+}
 ```
 
 ## Arguments Reference
--- a/docs/data-sources/azapi_resource_action.md
+++ b/docs/data-sources/azapi_resource_action.md
@@ -50,7 +50,24 @@ data "azapi_resource_action" "example" {
 }
 ```
 
-Here's an example to use `azapi_resource_action` resource to perform a provider action.
+Here's an example to use the `azapi_resource_action` data source to get a provider's permissions.
+
+```hcl
+provider "azurerm" {
+  features {}
+}
+
+data "azurerm_client_config" "current" {}
+
+data "azapi_resource_action" "test" {
+  type        = "Microsoft.Resources/providers@2021-04-01"
+  resource_id = "/subscriptions/${data.azurerm_client_config.current.subscription_id}/providers/Microsoft.Network"
+  action      = "providerPermissions"
+  method      = "GET"
+}
+```
+
+Here's an example to use the `azapi_resource_action` data source to perform a provider action.
 
 ```hcl
 terraform {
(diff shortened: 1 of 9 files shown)


internal/services/parse/resource.go
```diff
@@ -46,6 +46,13 @@ func NewResourceID(name, parentId, resourceType string) (ResourceId, error) {
 			azureResourceId = fmt.Sprintf("/subscriptions/%s", name)
 		case arm.TenantResourceType.String():
 			azureResourceId = "/"
+		case arm.ProviderResourceType.String():
```
magodo, line 49: Persumably this wants to change? Simply merging into the `default` branch and check whether there is only one segment in the `azureResourceType` should suffice?

docs/resources/azapi_resource_action.md
```diff
@@ -69,6 +69,28 @@ resource "azapi_resource_action" "stop" {
...
+terraform {
+  required_providers {
+    azapi = {
+      source = "Azure/azapi"
+    }
+  }
+}
+
+data "azapi_resource_action" "test" {
+  type        = "Microsoft.Resources/providers@2020-04-01-preview"
```
magodo, line 84: Shouldn't the `type` here be `Microsoft.ResourceGraph@2020-04-01-preview`? The `Microsoft.Resources/providers` has no API version `2020-04-01-preview`, which seems only map to a small set of operations that are defined in the `resources` RP, e.g. https://github.com/Azure/azure-rest-api-specs/blob/e680f9bf4f154ec427ba7feac432ed8efd9665d0/specification/resources/resource-manager/Microsoft.Resources/stable/2022-09-01/resources.json#L2131.

On the other hand, the matching resource type and its API version is essential for the language server?

docs/data-sources/azapi_resource_action.md
```diff
@@ -50,6 +50,25 @@ data "azapi_resource_action" "example" {
...
+terraform {
+  required_providers {
+    azapi = {
+      source = "Azure/azapi"
+    }
+  }
+}
+
+resource "azapi_resource_action" "test" {
+  type        = "Microsoft.Resources/providers@2021-04-01"
```
magodo, line 65: Same here, this wants to be `Microsoft.Resource@2021-04-01`

docs/guides/frequently_asked_questions.md
```diff
@@ -55,3 +55,13 @@ its request body won't contain any subnets definitions, so existing subnets will
...
+
+## What are the types for the resource group, subscription and tenant?
+
+| Resource Type | Type | Example |
+| --- | --- | --- |
+| resource group | `Microsoft.Resources/resourceGroups` | `/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg` |
+| subscription | `Microsoft.Resources/subscriptions` | `/subscriptions/00000000-0000-0000-0000-000000000000` |
+| tenant | `Microsoft.Resources/tenants` | `/tenants/00000000-0000-0000-0000-000000000000` |
+| management group | `Microsoft.Management/managementGroups` | `/providers/Microsoft.Management/managementGroups/00000000-0000-0000-0000-000000000000` |
+| provider | `Microsoft.Resources/providers` | `/providers/Microsoft.Storage` |
```
magodo, line 67: Can we add an additional column named `Comment`, and change it to something like below (probably with some different wordings):

```suggestion
| generic provider | `Microsoft.Resources/providers` | `<scope>/providers/<RP>` | This is the generic provider resource type whose operations are defined by ARM, and is supported by all of Azure RPs (e.g. actions like `register`, `providerPermissions`, etc.) |
| concrete provider | `Microsoft.Resources/providers` | `<scope>/providers/Microsoft.Storage` | This is the concrete provider whose operations are only defined for this provider itself |
```

magodo (approved):

docs/guides/frequently_asked_questions.md
ms-henglu: Great idea! I'll add it.

PR #332, opened 2023-08-10
add examples

ms-henglu:
These examples are converted from `azurerm` test cases. They're verified test cases.

Previous examples are moved to `quickstarts` folder, and duplicated examples are removed.


magodo (approved):
Good job :+1:

PR #336, opened 2023-08-15
bugfix: `id` is not stored when error happens during creation

ms-henglu:


internal/services/azapi_resource.go
```diff
@@ -408,6 +408,11 @@ func resourceAzApiResourceCreateUpdate(d *schema.ResourceData, meta interface{})
 
 	_, err = client.CreateOrUpdate(ctx, id.AzureResourceId, id.ApiVersion, body)
 	if err != nil {
+		if d.IsNewResource() {
+			if _, err := client.Get(ctx, id.AzureResourceId, id.ApiVersion); err == nil {
```
magodo, line 412: It is mostly fine, except there might be some resource that you have permission to create, but no permission to read (e.g. key vault key with only write policy?). In this case, you'll regard the resource doesn't exist, whilst it does exist. So probably we shall either check the error for status code further, or we can split the `client.CreateOrUpdate`?

magodo (commented):
Thanks for this PR!
I have left some additional mostly minor comments that once addressed this should be good to merge 👍

internal/services/azapi_resource.go
ms-henglu: Good point! But I think I'll skip this case, because if the user has no permission to read the resource, the `azapi_resource` does not work because it always calls the GET.

magodo: Fair enough :)

Issue #271, opened 2023-03-18
Failed GET before PUT for azapi_resource_action


daveneeley:
The following code block appears to call a GET instead of a PUT when the resource in question is created in the same execution plan.

```
resource "azapi_resource_action" "tf_set_keys" {
  type                   = "Microsoft.Web/sites@2022-03-01"
  resource_id            = azurerm_windows_function_app.tf.id
  action                 = "host/default/functionKeys/tf_key"
  method = "PUT"
  response_export_values = ["*"]
  body = jsonencode({
    properties = {
      name = "tf_key"
      value = random_password.tf_key.result
    }
  })
}
```
[The full script](https://gist.github.com/daveneeley/bd69228cb8ae9f848c0295a5973805a6) will FAIL the first time through with a 405 error, but PASS the second time with no changes. Using the `time_sleep` module doesn't affect the behavior.

```
// Error: performing action host/default/functionKeys/tf_key of  "Resource: (ResourceId \"/subscriptions/redacted/resourceGroups/redacted/providers/Microsoft.Web/sites/redacted/slots/staging\" / Api Version \"2022-03-01\")": GET https://management.azure.com/subscriptions/redacted/resourceGroups/redacted/providers/Microsoft.Web/sites/redacted/slots/staging/host/default/functionKeys/tf_key
│ --------------------------------------------------------------------------------
│ RESPONSE 405: 405 Method Not Allowed
│ ERROR CODE UNAVAILABLE
│ --------------------------------------------------------------------------------
│ <Error><Message>The requested resource does not support http method 'GET'.</Message></Error>
│ --------------------------------------------------------------------------------
│ 
│ 
│   with azapi_resource_action.tf_set_keys,
│   on main.tf line 155, in resource "azapi_resource_action" "tf_set_keys":
│  155: resource "azapi_resource_action" "tf_set_keys" {
```

Issue #184, opened 2022-09-14
Error trying to implement cost management exports for billing account scope


jrgrazina:
**Description**

 I'd like to be able to create a cost management export with the scope being the billing account. We use enterprise enrolment for Azure and this would be super useful for us.
 
I've attempted to use the azapi. Whilst it works (creates the Exports at a Billing Account Scope) it won't complete in TF and then write to the state:
_this_

c6713524e98/providers/Microsoft.CostManagementExports]
module.azure_interno.azapi_resource.cm_export: Creating...
module.azure_interno.azapi_resource.cm_export: Still creating... [10s elapsed]
module.azure_interno.azapi_resource.cm_export: Still creating... [20s elapsed]
module.azure_interno.azapi_resource.cm_export: Still creating... [30s elapsed]
module.azure_interno.azapi_resource.cm_export: Still creating... [40s elapsed]
module.azure_interno.azapi_resource.cm_export: Still creating... [50s elapsed]
module.azure_interno.azapi_resource.cm_export: Still creating... [1m0s elapsed]
╷
│ Error: creating/updating "Resource: (ResourceId \"/providers/Microsoft.Billing/billingAccounts/<BILLING_ACCT_#>/providers/Microsoft.CostManagement/exports/finopsreportnosmonthly\" / Api Version \"2021-10-01\")": Get "https://costmanagement.trafficmanager.net/providers/Microsoft.Billing/billingAccounts/<BILLING_ACCT_#>/providers/Microsoft.CostManagement/exports/finopsreportmonthly?api-version=2021-10-01": local error: tls: no renegotiation
│ 
│   with module.azure_interno.azapi_resource.cm_export,
│   on model_integration/main.tf line 81, in resource "azapi_resource" "cm_export":
│   81: resource "azapi_resource" "cm_export" {
│ 
╵
 
 _and that is the model used on azapi provider_
```
  # module.azure_interno.azapi_resource.cm_export will be created
  + resource "azapi_resource" "cm_export" {
      + body                      = (known after apply)
      + id                        = (known after apply)
      + ignore_casing             = true
      + ignore_missing_property   = true
      + location                  = (known after apply)
      + name                      = "finopsreportmonthly"
      + output                    = (known after apply)
      + parent_id                 = "/providers/Microsoft.Billing/billingAccounts/<BILLING_ACCT_#>"
      + response_export_values    = [
          + "*",
        ]
      + schema_validation_enabled = false
      + tags                      = (known after apply)
[...]

PR #341, opened 2023-08-17
ignore polling error from nonstandard LRO APIs

ms-henglu:
fixes https://github.com/Azure/terraform-provider-azapi/issues/271, https://github.com/Azure/terraform-provider-azapi/issues/184


magodo (commented):
Both cases look tricky to me, for both cases, I'd prefer we introduce a new toggle to bypass polling, which looks more sensible?

magodo (approved):

PR #343, opened 2023-08-17
retry in case of empty response while fetching resource

harshavmb:
This PR is an additional `retry` function with `azapi_resource` data provider. 

The use case for this `retry` logic is that we have few Azure policies to update DNS configuration of `private endpoints` created by various users. There is some slight delay between the private endpoint resource creation & the Azure policy to create A-record & update the private endpoint DNS configuration. We have seen the time lag of up to 30-40 seconds. 

Currently, we are using either `null_resource` & `sleep` providers. We wait deliberately for certain time expecting the configuration to be updated. Nevertheless, it's not a clean solution. We create inherent dependency on `az-cli`.

With this `retry` logic, `azapi_resource` data provider retries until non empty response is returned by the Azure API. 

I know retrying bluntly on empty response isn't wise but I'm open to your suggestions.

### DIFF
--- a/internal/services/azapi_resource_data_source.go
+++ b/internal/services/azapi_resource_data_source.go
@@ -14,6 +14,7 @@ import (
 	"github.com/Azure/terraform-provider-azapi/internal/services/validate"
 	"github.com/Azure/terraform-provider-azapi/internal/tf"
 	"github.com/Azure/terraform-provider-azapi/utils"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
@@ -103,6 +104,7 @@ func resourceAzApiDataSourceRead(d *schema.ResourceData, meta interface{}) error
 		id = buildId
 	}
 
+	var responseBody interface{}
 	responseBody, err := client.Get(ctx, id.AzureResourceId, id.ApiVersion)
 	if err != nil {
 		if utils.ResponseErrorWasNotFound(err) {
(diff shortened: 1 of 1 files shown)


harshavmb:
> @microsoft-github-policy-service agree

harshavmb:
Hi,

Any update on this PR?

ms-henglu:
Hi @harshavmb ,

Thank you so much for taking time to work on this feature! And apologize for the late response.

We'll think about this feature request and improve it.

ms-henglu:
We received a related feature request: https://github.com/Azure/terraform-provider-azapi/issues/372

harshavmb:
Hi @ms-henglu ,

I'm afraid the feature request you linked isn't the one this PR is about. It's not the resource, it's a data provider. 
As I explained in the PR description, the create part is handled by Azure policies which have few seconds of latency & with data providers we would like to retry until the ARM resource is created by the policy.

matt-FFFFFF:
Although, it might be a different scenario if the API does not return an error (which, I think it should)

harshavmb:
Hi @matt-FFFFFF ,

Indeed if you prefer to have a common function, both resources & data sources could use that but I would like to have this available with data sources too as we have different use-cases than the one mentioned in #372.

MatLambethMS:
Hi folks. How is this one looking? This feature would be extremely useful and looks like a lot of the hard work has been done.

ms-henglu:
Hi @harshavmb ,

Thank you for taking time to open this PR, but the would be solved by https://github.com/Azure/terraform-provider-azapi/pull/392 which supports customized retry error, please subscribe https://github.com/Azure/terraform-provider-azapi/pull/392 for latest updates.

I'll close this PR, but feel free to reopen it if there's any question.

PR #350, opened 2023-09-15
add a live traffic logger

ms-henglu:
Add a new logger to record the traffic in a structured way, so it would be much easier to parse the traffic from the logs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
-## v1.10.0
+## v1.10.0 (unreleased)
 ENHANCEMENTS:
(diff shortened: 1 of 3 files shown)


magodo (commented):
Overall it LGTM, while does it mean then we will have duplicated logs for both the original unstructured manner and this strucutred one?

ms-henglu:
> Overall it LGTM, while does it mean then we will have duplicated logs for both the original unstructured manner and this strucutred one?

Yes, this structured logger is mainly used for Armstrong integration. I'll think about retiring the original one in a major version to avoid breaking changes.

magodo (approved):

PR #361, opened 2023-10-16
`azapi` provider supports `endpoint`

ms-henglu:
This PR makes azapi provider to support customized endpoint like login endpoint, resource manager endpoint

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -78,6 +78,18 @@ It's possible to configure the behaviour of certain resources using the followin
 
 * `default_naming_suffix` - (Optional) The default name suffix to create the azure resource. Used together with `name` in each resource block. Conflicts with `default_name`. Changing this forces new resources to be created.
 
+* `endpoint` - (Optional) A `endpoint` block as defined below.
+
+---
+
+A `endpoint` block supports the following:
+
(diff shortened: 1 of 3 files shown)


docs/index.md
```diff
@@ -78,6 +78,18 @@ It's possible to configure the behaviour of certain resources using the followin
 
 * `default_naming_suffix` - (Optional) The default name suffix to create the azure resource. Used together with `name` in each resource block. Conflicts with `default_name`. Changing this forces new resources to be created.
 
+* `endpoint` - (Optional) A `endpoint` block as defined below.
```
magodo, line 81: The naming here is bit of confused, shall we change them to align with the structure of the Go SDK?

magodo (changes requested):

docs/index.md
ms-henglu: It's aligned with azure cli: https://learn.microsoft.com/en-us/cli/azure/manage-clouds-azure-cli#list-available-clouds
and raw response from the metadata url.

magodo: I've tried this https://management.azure.com/metadata/endpoints?api-version=2022-09-01, which has different structure than the CLI one, which I felt confused..

magodo (approved):

PR #368, opened 2023-11-01
`azapi_resource_action` - supports GET method

ms-henglu:


magodo (commented):
Wondering the motivation of this given we already have `azapi_resource_action` DS.

ms-henglu:
> Wondering the motivation of this given we already have `azapi_resource_action` DS.

It will be used in Armstrong for ARM review. There're GET APIs that couldn't be called like data source, so we need this feature to make the API more flexible.

magodo:
@ms-henglu Could you elaborate more about the case that *GET APIs that couldn't be called like data source*, somewhere in the data source document as a limitation and guide users to use the resource instead?

magodo:
By offline discussion, this feature is for some obscure internal use case, which most likely won't be encountered by users. So extending the `azapi_resource` operation type to include `GET` shall be added, but not documented to be known/used by regular users.

PR #384, opened 2023-12-08
feat: support for `when` on resource_action

### LINKED ISSUE #314: Resource `azapi_resource_action` supports `when=destroy`
Resource `azapi_resource_action` supports `when=destroy`

Add a new field `when` to the resource `azapi_resource_action` to allow users to specify when to run the action. The default value is `create`. If the value is `destroy`, the action will be run when the resource is destroyed.

**Example 1**

This function could be used to do some clean up work when the resource is destroyed. In the following config, when running `terraform destroy`, the action `stop_vm` will be run to stop the virtual machine before the resource is destroyed.

```hcl
resource "azapi_resource" "vm" {
  type = "Microsoft.Compute/virtualMachines@2020-06-01"
  ...
}

resource "azapi_resource_action" "stop_vm" {
  resource_id = azapi_resource.vm.id
  action_name = "stop"
  when = "destroy"
}
```

**Example 2**

This function could be used to support the azure resource which doesn't follow the standard CRUD guideline.

```hcl
// POST /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.RP/workspaces/default/createFoo
resource "azapi_resource" "create_foo" {
  type = "Microsoft.RP/workspaces@2020-06-01"
  resource_id = "${local.rg.id}/providers/Microsoft.RP/workspaces/default"
  action = "createFoo"
  body = ...
  method = "POST"
}

// DELETE /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.RP/workspaces/default/deleteFoo
resource "azapi_resource_action" "delete_foo" {
  type = "Microsoft.RP/workspaces@2020-06-01"
  resource_id = "${local.rg.id}/providers/Microsoft.RP/workspaces/default"
  action = "deleteFoo"
  when = "destroy"
  body = ...
}

```

matt-FFFFFF:
Implements `when` property for `azapi_resource_action`

fixes #314

### DIFF
--- a/docs/resources/azapi_resource_action.md
+++ b/docs/resources/azapi_resource_action.md
@@ -8,11 +8,11 @@ description: |-
 
 # azapi_resource_action
 
-This resource can perform any Azure resource manager resource action. 
+This resource can perform any Azure resource manager resource action.
 It's recommended to use `azapi_resource_action` resource to perform actions which change a resource's state, please use `azapi_resource_action` data source,
 if user wants to perform readonly action.
 
--> **Note** When delete `azapi_resource_action`, no operation will be performed.
+-> **Note** The action can be performed on either apply or destroy. The default is apply, see `when` argument for more details.
 
 ## Example Usage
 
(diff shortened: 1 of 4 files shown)


matt-FFFFFF:
Requesting review from @ms-henglu

ms-henglu (changes requested):
Thanks for taking time to work on this PR!

This PR looks good to me!

But one minor suggestion is that would you please add the new field to the doc: https://github.com/Azure/terraform-provider-azapi/blob/main/docs/resources/azapi_resource_action.md

matt-FFFFFF:
@ms-henglu have updated docs in 4e7c4a8

ms-henglu (approved):

PR #392, opened 2024-01-02
feat: user defined retryable errors

### LINKED ISSUE #372: feature request: user configurable retryable errors
Hi,

I would like to propose that the resources contained within this provider have an additional schema property to identify retryable errors.

With any complex cloud infrastructure, retryable errors are a fact of life. We could make our customers' lives easier by allowing them (and module authors) the ability to configure retries.

Inspiration: https://pkg.go.dev/github.com/cenkalti/backoff

## Proposed configuration

```terraform
  # usual resource properties

  # Retry configuration
  retry {
    interval_seconds     = 5    # mandatory
    randomization_factor = 0.5  # optional, default is 0.5
    multiplier           = 2    # optional, default is 2
    error_message_regex  = [
      "my error message",
      "end of error message$"
    ]
  }

  # Retry will continue up to configured timeouts
  timeouts {
    create = "10m"
  }
}
```

matt-FFFFFF:
Adds retryable errors for all non-data plane resources and data sources.

* resource_client updated with `Requester` interface to account for retry config and mocking
* resource schema and methods updated
* Mocks resource client added for testing
* Tests added for resource client with retry

Waiting on #505 to see if I need to manually update docs or not

@ms-henglu the vendoring was inconsistent and I don't think vending deps is recommended by 1ES, hence removed. Please let me know if you want to add back?

Issues:

fixes #372
fixes #343


internal/acceptance/testcase.go
```diff
@@ -123,10 +123,10 @@ func (td TestData) externalProviders() map[string]resource.ExternalProvider {
 
 func PreCheck(t *testing.T) {
 	variables := []string{
-		"ARM_CLIENT_ID",
```
harshavmb, line 126: Hi @matt-FFFFFF ,

I know this is still a draft but thought of letting you know that this is a breaking change. The `NewDefaultAzureCredential` function part of `azidentity` sdk uses env vars with `AZURE_*` as documented [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/README.md?plain=1#L170-L172). IMO `ARM_*` vars should still be supported as they are heavily used with `azurerm` provider.

matt-FFFFFF: The use of `NewDefaultAzureCredential` means that you can test locally when logged in with az cli, or even use managed service identity in your self-hosted runner. Nonr of these methods require these environment variables to be set.

The removal of `ARM_CLIENT_ID` from the pre-check means that you do not _have_ to supply it but can still do so if you are using an authentication mechanism that requires them.

The go SDK will provide you a meaningful error if it cannot authenticate.

harshavmb: okay, as I understood from you is that it's just internal testing of the provider. Nothing changes for the end user.

internal/services/azapi_resource_data_source.go
```diff
@@ -188,7 +192,18 @@ func (r *AzapiResourceDataSource) Read(ctx context.Context, request datasource.R
 		id = buildId
 	}
 
-	client := r.ProviderData.ResourceClient
+	var client clients.Requester
+	client = r.ProviderData.ResourceClient
+	if !model.RetryableErrors.IsNull() && !model.RetryableErrors.IsUnknown() {
+		bkof, regexps := clients.NewRetryableErrors(
+			model.RetryableErrors.GetIntervalSeconds(),
```
ms-henglu, line 199: The default value seem doesn't work on a data source. So the inverval second becomes 0 when using the retry feature with a data source.

internal/services/azapi_resource_data_source.go
```diff
@@ -188,7 +192,18 @@ func (r *AzapiResourceDataSource) Read(ctx context.Context, request datasource.R
...
+			model.RetryableErrors.GetMaxIntervalSeconds(),
+			model.RetryableErrors.GetMultiplier(),
+			model.RetryableErrors.GetRandomizationFactor(),
```
ms-henglu, line 202: Same as above, because of the default value is not set, block like the following in a data source will cause panic:

```hcl
data "azapi_resource" "test" {
  type = "Microsoft.Resources/resourceGroups@2024-03-01"
  name = "henglu82"
  retryable_errors = {
    error_message_regex = [ "ResourceGroupNotFound" ]
  }
}
```

internal/retry/schema.go
```diff
@@ -0,0 +1,84 @@
...
+
+const (
+	intervalSecondsAttributeName     = "interval_seconds"
+	maxIntervalSecondsAttributeName  = "max_interval_seconds"
+	multiplierAttributeName          = "multiplier"
+	randomizationFactorAttributeName = "randomization_factor"
+	errorMessageRegexAttributeName   = "error_message_regex"
+)
+
+func SingleNestedAttribute(ctx context.Context) schema.Attribute {
```
ms-henglu, line 25: How about renaming the block to `retry`? Because maybe in the future we could introduce other customized retry logics like retry untils the resource's status becomes X.

```
retry {
    error_message_regex = [ "ResourceGroupNotFound" ]
    multiplier = 1.5
    randomization_factor = 0.5
    interval_seconds = 10
    max_interval_seconds = 180
}
```

ms-henglu (changes requested):
Thanks for the PR! It mostly LGTM, but I found some issues when I test it, please check.

ms-henglu:
Hi @matt-FFFFFF ,

I use the blow config to test the retry logic, but I found the requests are not sent by the configured interval.

```hcl
resource "time_sleep" "wait_30_seconds" {
  create_duration = "30s"
}

resource "azapi_resource" "resourceGroup" {
  type      = "Microsoft.Resources/resourceGroups@2021-04-01"
  name      = "henglu82"
  location  = "westus"
  body     = {}  # Resource group does not require additional properties in the body
  lifecycle {
    ignore_changes = [ tags ]
  }
  depends_on = [ time_sleep.wait_30_seconds ]
}


data "azapi_resource" "test" {
  type = "Microsoft.Resources/resourceGroups@2024-03-01"
  name = "henglu82"
  retryable_errors = {
    error_message_regex = [ "ResourceGroupNotFound" ]
    multiplier = 1.5
    randomization_factor = 0.5
    interval_seconds = 10
    max_interval_seconds = 180
  }
}
```


image

matt-FFFFFF:
Thank you @ms-henglu !

I am back on Tuesday when I will work on these fixes

matt-FFFFFF (commented):
Hey @ms-henglu

I have better understood how the exponential backoff works. I have added a backoff.Reset() call to ensure the timers are reset and then I've added a test to validate the exponential backoff.

[// TestRetryClient tests the retry client with a mock client that returns an error.](https://github.com/matt-FFFFFF/terraform-provider-azapi/blob/a32f59d8fbbe8aba665aa206368a3020c7e749f1/internal/clients/resource_client_test.go#L13-L46)

Please see this new test

internal/retry/schema.go
matt-FFFFFF: Refactored in: 4f6a1361e1b83f1e51f6524f70ad2a2515ba2ce8

internal/services/azapi_resource_data_source.go
matt-FFFFFF: Fixed in cb47241e2e5de914d6f74a3d1beda911dc950db9

matt-FFFFFF:
@ms-henglu I believe I have addressed the issues you found - please can you review?

Thank you

ms-henglu (approved):
Thanks for the PR! LGTM

ms-henglu:
I'll open another PR to update the generated docs.

Issue #417, opened 2024-02-21
Bug report: the `name` field in the `body` is removed in request body when using `ignore_body_changes`


ms-henglu:
Context: Updating the "Microsoft.NotificationHubs/namespaces@2023-10-01-preview" resource requires the `name` field to be set. But when using `ignore_body_changes` future, the combined body will be processed to remove the readonly fields based on the bicep-types. And the `name` field is both required and deployed time constant, which should be kept in the request body.

Actual behavior: The `name` field is removed in the request body.

Expected behavior: The `name` field should be kept in the request body.

PR #418, opened 2024-02-21
bugfix: `azapi_resource` removes some required fields in the `body` when using `ignore_body_changes` feature

ms-henglu:
fixes https://github.com/Azure/terraform-provider-azapi/issues/417


magodo (approved):
LGTM!

PR #434, opened 2024-03-14
bugfix: timeout is not set when import `azapi_resource`

ms-henglu:
This PR sets a default null value for the timeouts when import the resource. If not set, a value conversion error will happen.


magodo (approved):
Do you think this is a bug in the fw?

Issue #497, opened 2024-05-06
Schema validation fails when using azapi_resource for keyvault with random string in the name


bengodeloson:
We are currently using azapi_resource for our keyvault. Upon upgrading the provider from 1.12.1 to 1.13.0 and 1.13.1, we encounter an issue which has this error:

> ##[error][TerraformClient]: Terraform command finished with errors: ╷
> │ Error: Invalid configuration
> │
> │   with module.keyvault.azapi_resource.key_vault,
> │   on tempSource/akv.tf line 5, in resource "azapi_resource" "key_vault":
> │    5: resource "azapi_resource" "key_vault" {
> │
> │ embedded schema validation failed: the argument "body" is invalid:
> │ `name` is invalid, string does not match pattern ^[a-zA-Z0-9-]{3,24}$
> │ You can try to update `azapi` provider to the latest version or disable the validation using the feature flag `schema_validation_enabled = false` within the resource block
> ╵

We are adding random string in our keyvault name which has the following configuration:
```
resource "random_string" "akv_suffix" {
  length  = 3
  special = false
  upper   = false
}

resource "azapi_resource" "key_vault" {
  type      = "Microsoft.KeyVault/vaults@2022-07-01"
  name      = "akv${var.name}${random_string.akv_suffix.result}"
  location  = var.location
  parent_id = local.resource_group_id
  tags      = local.tags
...
}
```

When we remove the ${random_string.akv_suffix.result} or disabling the schema validation, the plan stage succeeds. We really want to add the schema validation to our module and this was working previously in 1.12.1. Can anyone help us with this issue?

ms-henglu:
Hi @bengodeloson ,

Thank you for taking time to report this issue!

I'll fix it in the next release. As a workaround, please disable the scheme validation and you could enable it after the bug fix is released.

ms-henglu:
Hi @bengodeloson 

The bugfix has been released in the azapi v1.14.0. Please give it a try.

ms-henglu:
Close this issue as it's completed, but feel free to reopen it if there's any question.

bengodeloson:
Hi @ms-henglu ,

I've validated the fix and it works. Thank you.

PR #509, opened 2024-05-20
bugfix: fix validation for unknown string value

ms-henglu:
related issue: https://github.com/Azure/terraform-provider-azapi/issues/497


magodo:
Can we add more comment to clarify the behavior here when the value being validated is an unknown value?

magodo (approved):

PR #510, opened 2024-05-21
`azapi_*` - support for the `timeouts.update` field

ms-henglu:
This PR adds `timeouts.update` field which is removed in v1.13.x by mistake.

To avoid potential breaking changes introduced by this PR, I've added more tests:

1. Upgrade from v1.12.1 to current version, the updated migration function will update the schema from v0 to v1 correctly.
2. Upgrade from v1.13.1 to current version, the schema is already updated to v0 to v1. But in v1.13.x, the update timeout is not supported, so upgrading will have no breaking changes. And users could still be able to add update timeout after using the latest version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## v1.14.0 (unreleased)
 
+ENHANCEMENTS:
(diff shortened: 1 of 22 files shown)


magodo:
LGTM :+1:

magodo (approved):

Issue #523, opened 2024-06-03
Add extension parentID type for billingAccount


PPACI:
Hello
The following code doesn't support extension on anything else than resource group.
https://github.com/Azure/terraform-provider-azapi/blob/e53a2e48d676da9d01efea6eb240736f7d6374bc/internal/services/parse/resource.go#L193-L202

this means it would not be possible to create [costallocationrules](https://learn.microsoft.com/en-us/azure/templates/microsoft.costmanagement/costallocationrules?pivots=deployment-language-terraform) since it's a billingAccount extension. It seems we can't have the parentId right, there is no way to get a parentId of type extension on a billingAccount.

I found the following workaround by abusing the validation code, but it's not pretty...
Would it be possible to add other extension type to support this use case?
thank you very much!

That create the resource, but it can't be managed afterward.

But beside this (edge case I agree), thank you for the great work on this project!!

```go
resource "azapi_resource" "cost_allocation_rules" {
  for_each  = { for rule in var.cost_allocation_rules_resource_group : rule.sourceResourceGroupName => rule }
  type      = "costAllocationRules@2023-11-01" # As long as it's doesn't contain "/" it's fine
  name      = "Microsoft.CostManagement/costAllocationRules/${each.value["sourceResourceGroupName"]}-to-${data.azurerm_subscription.target.display_name}"
  parent_id = "/providers/Microsoft.Billing/billingAccounts/${var.billing_account_id}"

  /*
  case !strings.Contains(azureResourceType, "/"):
		// case 0: resource type is a provider type
		// avoid duplicated `/` if parent_id is tenant scope
		scopeId := parentId
		if parentId == "/" {
			scopeId = ""
		}
		azureResourceId = fmt.Sprintf("%s/providers/%s", scopeId, name)
  

but for this to work we have to pass an arbritary resource type
  and thus disable schema validation
  */
  schema_validation_enabled = false

  body = jsonencode({...})
}
```

ms-henglu:
Hi @PPACI ,

Thank you for taking time to report this!

I'll open a PR to fix it shortly.

PPACI:
Amazing to hear! thank you very much.

PR #526, opened 2024-06-05
bugfix: extension resource could be applied to any scope

ms-henglu:
fixes https://github.com/Azure/terraform-provider-azapi/issues/523


PPACI:
The test you've added also act as a great documentation for extension resource. thank you very much!

Do you have any estimate on when we could expect this to be merged? Just to know if we're speaking weeks or month here.

magodo (approved):

ms-henglu:
Thanks @PPACI , I think it'll be released before end of June.

Issue #534, opened 2024-06-19
Fallback for OIDC


rdvansloten:
When setting `use_oidc` to `true` (required to use OIDC in GitHub Workflows), local builds don't fall back like the AzureRM provider does.

For example, when setting OIDC in the azurerm backend:

```hcl
backend "azurerm" {
    resource_group_name  = "***"
    storage_account_name = "***"
    container_name       = "tfstate"
    key                  = "***/tfstate"
    subscription_id      = "***"
    tenant_id            = "***"
    use_oidc             = true
    client_id            = "***"
    use_azuread_auth     = true
  }
```

This will gracefully fall back to Azure CLI auth when running the config locally on a workstation. However,

```hcl
provider "azapi" {
  use_oidc = true
}
```

Does not:

```sh
/ Api Version "2023-05-02-preview"): ChainedTokenCredential: failed to acquire a token.
 Attempted credentials:
      ClientAssertionCredential: unable to resolve an endpoint: http
call(https://login.microsoftonline.com//v2.0/.well-known/openid-configuration)(GET) error: reply status code was
400:
{"error":"invalid_tenant","error_description":"AADSTS90002: Tenant 'v2.0' not found. Check to make sure you have the
correct tenant ID and are signing into the correct cloud. Check with your subscription administrator, this may
happen if there are no active subscriptions for the tenant. Trace ID: 9b68a459-654e-438b-a857-78f951622300
Correlation ID: 1ea4c968-6341-4bd3-ba14-892a54681023 Timestamp: 2024-06-19
07:33:42Z","error_codes":[90002],"timestamp":"2024-06-19
07:33:42Z","trace_id":"9b68a459-654e-438b-a857-78f951622300","correlation_id":"1ea4c968-6341-4bd3-ba14-892a54681023","error_uri":"https://login.microsoftonline.com/error?code=90002"}
```

Removing `use_oidc` locally works, but is not ideal, since this of course triggers changes in Git and might accidentally be pushed back to repo and ruin the test in Workflows.

ms-henglu:
Hi @rdvansloten ,

Thank you for taking time to report this issue!

I've opened a PR to fix it: https://github.com/Azure/terraform-provider-azapi/pull/536

And I'd like to double confirm that when use the config locally, there's no other OIDC related fields except `use_oidc=true` are used, right?

rdvansloten:
> Hi @rdvansloten ,
> 
> 
> 
> Thank you for taking time to report this issue!
[...]

PR #536, opened 2024-06-20
bugfix: oidc should not be used when required fields are missing

ms-henglu:
fixes https://github.com/Azure/terraform-provider-azapi/issues/534

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -832,6 +832,9 @@ func buildOidcCredential(model providerData, options azidentity.DefaultAzureCred
 	if err != nil {
 		return nil, err
 	}
+	if model.OIDCToken.ValueString() == "" && model.GetOIDCTokenFilePath() == "" && model.OIDCRequestToken.ValueString() == "" && model.OIDCRequestURL.ValueString() == "" {
+		return nil, fmt.Errorf("missing required OIDC configuration")
+	}
 	o := &OidcCredentialOptions{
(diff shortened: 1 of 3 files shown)


internal/provider/provider.go
```diff
@@ -832,6 +832,9 @@ func buildOidcCredential(model providerData, options azidentity.DefaultAzureCred
 	if err != nil {
 		return nil, err
 	}
+	if model.OIDCToken.ValueString() == "" && model.GetOIDCTokenFilePath() == "" && model.OIDCRequestToken.ValueString() == "" && model.OIDCRequestURL.ValueString() == "" {
```
ms-zhenhua, line 835: should be `||`, not `&&`?

ms-zhenhua (changes requested):

internal/provider/provider.go
ms-henglu: No, OIDC doesn't require all of them to be set.

ms-zhenhua, line 835:
```suggestion
	if model.OIDCToken.ValueString() == "" && model.GetOIDCTokenFilePath() == "" && (model.OIDCRequestToken.ValueString() == "" || model.OIDCRequestURL.ValueString() == "") {
```

ms-zhenhua (changes requested):

ms-zhenhua (approved):

PR #569, opened 2024-08-05
bugfix: resource definitions are not correctly cached

ms-henglu:


magodo:
Benchmark test
===

Testing with `terraform-client-go` for N rounds, where each round spawn the `azapi` provider, configure it, then loop N times to import and read a `azapi_resource`, whose RT is a virtual machine. Note that the actual API call in read/import is commented to ignore I/O delay.

Given N = 20,

### Before the Change

```
goos: linux
goarch: amd64
pkg: github.com/magodo/terraform-client-go/tfclient
cpu: 13th Gen Intel(R) Core(TM) i7-13800H
BenchmarkClient/Spawn/WithSchema-20                   20         713116134 ns/op
PASS
ok      github.com/magodo/terraform-client-go/tfclient  14.792s
go test -benchtime=20x -bench='BenchmarkClient/Spawn/WithSchema' .  13.39s user 1.31s system 96% cpu 15.203 total
```

After the change

```
goos: linux
goarch: amd64
pkg: github.com/magodo/terraform-client-go/tfclient
cpu: 13th Gen Intel(R) Core(TM) i7-13800H
BenchmarkClient/Spawn/WithSchema-20                   20         142416318 ns/op
PASS
ok      github.com/magodo/terraform-client-go/tfclient  2.981s
go test -benchtime=20x -bench='BenchmarkClient/Spawn/WithSchema' .  3.79s user 0.91s system 132% cpu 3.546 total
```

ms-henglu:
> Persumably we shall add a mutex in the `GetDefinition()` to avoid concurrent access to the `o.Definition`?

Makes sense. I've updated this PR as suggested.

magodo (approved):
LGTM, thanks!

PR #588, opened 2024-08-22
fix: always use non-retry client to check new resource does not exist

matt-FFFFFF:
Hi @ms-henglu 

In testing I found a scenario where we want to add a retryable error for a ResourceNotFound situation. For new resources, there is a check to ensure the resource does not already exist. If we use the retry client then the provider will infinite loop until the context deadline.

This fix ensures we use the non-retry client to check that the resource does not already exist!


ms-henglu (approved):
Thank you for this PR! LGTM

PR #592, opened 2024-08-27
bugfix: fix state migration for `response_export_values`

ms-henglu:
This PR fixes the state migration for `response_export_values`. Previously, the null list will be converted to a dynamic value whose underlying value is a null list. If the user doesn't specify the `response_export_values`, the config value should be a null dynamic value, and this will cause a plan diff because the structs are different.


ms-henglu:
cc @hqhqhqhqhqhqhqhqhqhqhq

magodo (approved):

PR #600, opened 2024-09-02
implement `replace_triggers_refs` for `azapi_resource`

hqhqhqhqhqhqhqhqhqhqhq:
for resource `azapi_resource`, the `replace_triggers_refs` keep a list of paths (JMES supported) pointing to values in the resource's **body**, when the value pointed by any one of the path is changed, the resource will be replaced.

example:
```
  type      = "Microsoft.Relay/namespaces@2021-11-01"
  parent_id = azurerm_resource_group.example.id
  name      = "xxx"
  location  = "westus"
  body = {
    properties = {
    }
    sku = {
      name = "Standard"
      tier = "Standard"
    }
  }

  replace_triggers_refs = ["sku"]
}
```

The current `sku` value is 
```
{
  name = "Standard"
  tier = "Standard"
}
```
if there is any changes above, the whole resource will be replaced.


internal/services/azapi_resource.go
```diff
@@ -196,6 +197,12 @@ func (r *AzapiResource) Schema(ctx context.Context, _ resource.SchemaRequest, re
 				},
 			},
 
+			"replace_triggers_refs": schema.ListAttribute{
```
ms-henglu, line 200: Please also implement this feature in the data plane resource.

ms-henglu, line 200: Please also implement this feature in the data plane resource.

internal/services/azapi_resource.go
```diff
@@ -499,6 +506,36 @@ func (r *AzapiResource) ModifyPlan(ctx context.Context, request resource.ModifyP
...
+		err = json.Unmarshal([]byte(plan.Body.String()), &data)
+		if err != nil {
+			response.Diagnostics.AddError("Invalid configuration", err.Error())
+			return
+		}
+		currentValues := flattenOutputJMES(data, refPaths)
+
+		// compare previous and current values
+		if !reflect.DeepEqual(previousValues, currentValues) {
+			response.RequiresReplace.Append(path.Root("body"))
```
ms-henglu, line 536: Is it possible to add the specific path like "body.properties.sku.name" that triggers the replacement?

ms-henglu (changes requested):
Thanks @hqhqhqhqhqhqhqhqhqhqhq for this PR! LGTM, just some minor suggestions, please check.

internal/services/azapi_resource.go
hqhqhqhqhqhqhqhqhqhqhq: I tried this, the cli will say:
```
Provider "registry.terraform.io/azure/azapi" has indicated "requires replacement" on azapi_resource.example for a
│ non-existent attribute path cty.Path{cty.GetAttrStep{Name:"body.properties.sku.name"}}.
```

Similar error message if I try with `body.sku.name`

internal/services/azapi_resource.go
hqhqhqhqhqhqhqhqhqhqhq: Got it, any need for other resources too?

ms-henglu: Thanks! No.

hqhqhqhqhqhqhqhqhqhqhq: Added & tested

ms-henglu (approved):

PR #621, opened 2024-09-19
fix: get after put

matt-FFFFFF:
Modifies retry client to test for status codes and callback funcs on the data interface{}.

CreateUpdate modified  to use this a specific retry client for the get that follows the put.

* Fixes #618

PR #627, opened 2024-09-27
avoid infinite retry

ms-henglu:
https://github.com/Azure/terraform-provider-azapi/pull/621 introduces logics that retry the GET 404 after PUT is successful.

This PR sets the maximum total time for retries. Because I found a case that the resource might not be successfully created, and the GET requests keeps failing.

image

### DIFF
--- a/internal/clients/data_plane_client.go
+++ b/internal/clients/data_plane_client.go
@@ -327,8 +327,8 @@ func (client *DataPlaneClient) Action(ctx context.Context, resourceID string, ac
 }
 
 func (retryclient *DataPlaneClientRetryableErrors) CreateOrUpdateThenPoll(ctx context.Context, id parse.DataPlaneResourceId, body interface{}, options RequestOptions) (interface{}, error) {
-	if retryclient.backoff == nil || len(retryclient.errors) == 0 {
-		return nil, fmt.Errorf("retry is not configured, please call WithRetry() first")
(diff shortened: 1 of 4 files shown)


ms-henglu:
Hi @matt-FFFFFF , please help review this PR, thanks!

matt-FFFFFF (commented):
Hi @ms-henglu 

The 404client should still respect the context and therefore the create timeout on the resource. Was this the case in your testing?

I can understand if we want a shorter maximum time for this get operation though.

We might want to consider making this max value configurable at a provider level to avoid having to release new version for edge cases?

ms-henglu:
Hi @matt-FFFFFF ,

Yes, it still respects the create timeout. But normally this will up to 30 minutes, so it's too long, it would be better to set a shorter timeout for this retry.

> We might want to consider making this max value configurable at a provider level to avoid having to release new version for edge cases?

This is a good idea, I guess there won't be many cases that user needs to configure it, I'll update this PR to support configuring this timeout by an environment variable.

matt-FFFFFF (approved):

magodo (approved):

matt-FFFFFF:
LGTM - do we want to document this environment variable?

ms-henglu:
> LGTM - do we want to document this environment variable?

Thanks! I think it's not necessary to document it because it'll not be commonly used.

PR #652, opened 2024-10-23
fix: Typo in README.md

DenisBalan:
Fix in terraform example reference for resource group. `example` should be used instead of `test`


ms-henglu (approved):
Thanks for the PR! LGTM

PR #653, opened 2024-10-25
Add unique_string function

smokedlinq:
This adds the ARM function `uniqueString` so that the community can migrate from existing ARM/BICEP based deployments that utilize the function to create resources.

The golang version was converted from a [C#/PowerShell utility](https://github.com/smokedlinq/powershell-az/blob/d5461aa62614ee741275b18ba7f710fb3e6b0856/powershell-az.psm1#L287) I wrote a few years ago.

The values can be verified with a simple bicep configuration such as:

```bicep
output value string = uniqueString('/subscriptions/00000000-0000-0000-0000-000000000000', 'resource-id')
```

P.S. Without a spreadable tfsdk type for a list of strings, an array seemed to be the most appropriate approach as the parameter.


internal/services/functions/unique_string_function.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	return builder.String()
+}
+
+func murmurHash64(value string) uint64 {
+	bytes := []byte(value)
+	return murmurHash64A(bytes, 0)
+}
+
+func murmurHash64A(data []byte, seed uint32) uint64 {
```
magodo, line 78: Can we add unit tests for this function as https://github.com/smokedlinq/powershell-az/blob/d5461aa62614ee741275b18ba7f710fb3e6b0856/powershell-az.tests.ps1?

smokedlinq: @magodo sorry for taking so long on this, but ya I added the same test cases from the PowerShell version.

ms-henglu (approved):

PR #657, opened 2024-11-01
feat: azapi_client_config - add object id

matt-FFFFFF:
This PR closes the gap to the azurerm equivalent by adding the `object_id` attribute to the `azapi_client_config` resource.

Also adds a convenience attribute to azapi_client_config to allow for adding the subscription resource id easily into `parent_id`.

@ms-henglu


internal/services/azapi_client_config_data_source.go
```diff
@@ -50,6 +52,15 @@ func (r *ClientConfigDataSource) Schema(ctx context.Context, request datasource.
...
 				Computed: true,
 			},
+
+			"subscription_resource_id": schema.StringAttribute{
+				Computed:            true,
+				MarkdownDescription: "The resource ID of the subscription. E.g. `/subscriptions/00000000-0000-0000-0000-000000000000`",
+			},
+
+			"object_id": schema.StringAttribute{
+				Computed: true,
```
ms-henglu, line 62: Please add a `MarkdownDescription` for this field.

internal/clients/client.go
```diff
@@ -86,7 +91,16 @@ func (client *Client) Build(ctx context.Context, o *Option) error {
 		"api-version",
 		"$skipToken",
 	}
-
+	tok, err := o.Cred.GetToken(client.StopContext, policy.TokenRequestOptions{
```
ms-henglu, line 94: Is it possible that we move this part to the ResourceManagerAccount's lazy init function `GetObjectId()`?
Because requesting token here will slow down the azapi initialization.

ms-henglu (changes requested):
Thanks @matt-FFFFFF for this PR. It mostly LGTM, just some minor comments, please check.

internal/services/azapi_client_config_data_source.go
matt-FFFFFF: Done :)

internal/clients/client.go
matt-FFFFFF: I have moved the token request to the GetObjectId() method. To do this I had to pass in the *Client as a dependency. Let me know what you think

matt-FFFFFF:
Hi @ms-henglu I have updated this and responded to your comments. Please can you review again? Thanks!

internal/clients/account.go
```diff
@@ -62,12 +73,63 @@ func (account *ResourceManagerAccount) GetSubscriptionId() string {
...
+	}
+
+	tok, err := account.client.Option.Cred.GetToken(account.client.StopContext, policy.TokenRequestOptions{
+		TenantID: account.client.Option.TenantId,
+		Scopes:   []string{account.client.Option.CloudCfg.Services[cloud.ResourceManager].Endpoint + "/.default"}})
+	if err != nil {
+		log.Printf("[DEBUG] Error getting requesting token from credentials: %s", err)
+	}
+
+	cl, err := parseTokenClaims(tok.Token)
```
ms-henglu, line 98: Do we need a nil-check on the `tok`? If above GetToken returns error, the `tok` could be nil here?

ms-henglu (changes requested):

internal/clients/account.go
matt-FFFFFF: `tok` is of type [AccessToken](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore@v1.13.0/internal/exported#AccessToken). This is a value type not a pointer.

The type declaration is:

```go
type AccessToken struct {
	Token     [string](https://pkg.go.dev/builtin#string)
	ExpiresOn [time](https://pkg.go.dev/time).[Time](https://pkg.go.dev/time#Time)
}
```

I refactored using a comparison of `tok.Token == ""` as this will be the default value if the token request fails.

WDYT?

[b7ae1bb](https://github.com/Azure/terraform-provider-azapi/pull/657/commits/b7ae1bb25967bce4ddf8a785ddf0d1255bc89e91)

ms-henglu (approved):
Thanks for the PR! LGTM

Issue #672, opened 2024-11-16
bug: using multiple `locks` can result in a deadlock


jaredfholgate:
If using multiple locks and the same locks are supplied to more than one resource, it can result in a deadlock due to the way the code handles the for loop for multiple locks. This only happens when timing is exact, but I have had it many times on a destroy run for vnet peering where I was attempting to lock both vnets for the forward and reverse peering.

For this code: 
https://github.com/Azure/terraform-provider-azapi/blob/0d963c24fd11d92345f410eed64af3fec6c22a58/internal/services/azapi_resource.go#L673

Here is an example:

- Resource 1
  - locks = [ “a”, “b” ]

- Resource 2
  - locks = [ “b”, “a” ]

1. Resource 1 locks “a”
1. Resource  2 locks “b”
1. Resource 1 tries to lock “b” and waits
1. Resource 2 tries to lock “a” and waits

This results in a deadlock, it waits forever and the apply never finishes. I left one running while I stepped away to sit an exam and it was still running after over 3 hours.

I can't immediately think of a solution to this, because adding a mutex before the `for` loop would block all resource with a lock and slow down the run.

Only other alternative is to support only one lock per resource or highlight this deadlock scenario in the docs to make people aware.

ms-henglu:
Hello @jaredfholgate ,

Thank you for taking time to report this issue!

I think I can fix it by sorting the locks, I'll open a PR to fix it in the next release. For users who use old versions, they need to use a fixed order of the locked resource IDs to avoid deadlock.

For example, always set the locks to `[a, b]` instead of `[b, a]`.

PR #676, opened 2024-11-19
bugfix: ordering locks to avoid deadlock

ms-henglu:
Fixes https://github.com/Azure/terraform-provider-azapi/issues/672


magodo (commented):
Isn't this a user error?

ms-henglu:
> Isn't this a user error?

Yes, it could be fixed by the user input. But I think the user might not understand why it's necessary to sort the locked resource IDs.

magodo (commented):
I'd suggest we change it to be a set if we want to reorder for the users, and mention this in the document.

ms-henglu:
Maybe not necessary to make it a breaking change, because we only use it in a sorted order.

magodo (approved):

Issue #630, opened 2024-09-29
Wrongly required update in-place for sorting identities array


jpenna:
It is updating the order of the `identity_ids` array, which I didn't change in the `.tf` file. Configuration:

```tf
resource "azapi_resource" "ai_service" {
  type      = "Microsoft.App/containerApps@2024-03-01"
  name      = "<NAME>"
  location  = module.consts.location
  parent_id = var.resource_group_id

  identity {
    type         = "UserAssigned"
    identity_ids = [var.acr_umi_id, azurerm_user_assigned_identity.kv_access_identity.id]
  }

  body = {
    properties = {
      environmentId = var.ca_env_id
      configuration = {
        activeRevisionsMode = "Single"

        registries = [{
          server   = var.acr_server
          identity = var.acr_umi_id
        }]
      }


      template = {
        terminationGracePeriodSeconds = 15

        scale = {
          minReplicas = 0
          maxReplicas = 1
        }

        serviceBinds = [{
          name      = "postgresql"
          serviceId = var.postgresql_app_id
        }]

        containers = [{
          name  = local.container_name
          image = "${var.acr_server}/ai_service:${var.rs_version}"
        }]
      }
    }
  }
}

``` 

## Plan output

Check `identify.identify_ids`:

```
    # module.ai_service_container_app.azapi_resource.ai_service will be updated in-place
  ~ resource "azapi_resource" "ai_service" {
      ~ body                      = {
          ~ properties = {
              ~ configuration = {
                  ~ ingress             = {
                      + transport  = "tcp"
                        # (3 unchanged attributes hidden)
                    }
                    # (3 unchanged attributes hidden)
                }
                # (2 unchanged attributes hidden)
            }
        }
        id                        = "/subscriptions/<SUBS_ID>/resourceGroups/<RG>/providers/Microsoft.App/containerApps/<NAME>"
        name                      = "<NAME>"
      ~ output                    = {} -> (known after apply)
        # (6 unchanged attributes hidden)

      ~ identity {
          ~ identity_ids = [
              - "/subscriptions/<SUBS_ID>/resourceGroups/<RG>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<KV_NAME>",
                "/subscriptions/<SUBS_ID>/resourceGroups/<RG>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<ACR_NAME>",
[...]

PR #679, opened 2024-11-25
Change identity_ids type to Set

Miouge1:
Change identity_ids type from List to Set.

The order of identity_ids is not consistent, therefore it should probably be a Set rather than a List.

See #630


Miouge1:

ms-henglu:
Thanks @Miouge1 for this PR! LGTM. But I need to run more tests to make sure this change won't break existing users.

PR #681, opened 2024-11-26
fix: improve retry behavior with resources at MG scope

matt-FFFFFF:
As per internal thread:

- Improved 403 handling for MG resources to include child resources
- Fix bug in get after put retry timeout so the environment variable works properly
- Added debug output to retry

I could also pass the regex to the get after put retry, but since the body is nil then the retry func catches the issue where we get a 403. WDYT @ms-henglu

### DIFF
--- a/internal/clients/data_plane_client.go
+++ b/internal/clients/data_plane_client.go
@@ -20,6 +20,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/terraform-provider-azapi/internal/services/parse"
 	"github.com/cenkalti/backoff/v4"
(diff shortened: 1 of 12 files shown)


ms-henglu:
Hi @matt-FFFFFF ,

Thank you for the PR, LGTM. 

> I could also pass the regex to the get after put retry, but since the body is nil then the retry func catches the issue where we get a 403. WDYT

I think it's better to allow the user specified `retry` block to control retry policy of the GET after PUT. Maybe combine both 404 or empty body retry with the user configured retry policy?

matt-FFFFFF:
Hi @ms-henglu I have rebased and updated this PR to pass the regex messages to the getafterput client.

WDYT?

ms-henglu (approved):
Thanks for the PR, LGTM.

Issue #711, opened 2025-01-08
[BUG] `azapi_*` cannot support query parameters of unknown type


cmaclaughlin:
I found this with `delete_query_parameters` of `azapi_resource` but since all query parameters introduced with #590 are the same "type", I am extrapolating the others would encounter similar issues.

I need to set a value based upon a variable. I have tried a few variations:

Attempt 1: Create local value and reference it 

(this is just 1 of the permutations I tried in terms of what is conditionally set)

```hcl
locals {
  hard_delete_parameter = {
    "forceToPurge" = tolist(["true"])
  }
  soft_delete_parameter = {
    "forceToPurge" = tolist(["false"])
  }

  default_delete_parameters = var.force_hard_delete ? local.hard_delete_parameter : local.soft_delete_parameter
}

resource "azapi_resource" "this" {
  delete_query_parameters = tomap(local.default_delete_parameters)
}
```

```bash
│ An unexpected error was encountered trying to build a value. This is always an error in the provider. Please report the following to the provider developer:
│ 
│ Received unknown value, however the target type cannot handle unknown values. Use the corresponding `types` package type or a custom type that handles unknown values.
│ 
│ Path: delete_query_parameters
│ Target Type: map[string][]string
│ Suggested Type: basetypes.MapValue
```

Attempts 2 & 3 Use string interpolation / inline conditional

```hcl
resource "azapi_resource" "this" {
  delete_query_parameters = {
    "forceToPurge" = tolist([ "${var.force_hard_delete}" ]) 
  }
}
```

```hcl
resource "azapi_resource" "this" {
  delete_query_parameters = {
    "forceToPurge" = tolist([ (var.force_hard_delete ? "true" : "false") ])
  }
}
```

```bash
 An unexpected error was encountered trying to build a value. This is always an error in the provider. Please report the following to the provider developer:
│ 
│ Received unknown value, however the target type cannot handle unknown values. Use the corresponding `types` package type or a custom type that handles unknown values.
│ 
│ Path: delete_query_parameters["forceToPurge"][0]
│ Target Type: string
│ Suggested Type: basetypes.StringValue
```

Finally, I do find success with the following

```hcl
resource "azapi_resource" "this" {
  delete_query_parameters = {
    "forceToPurge" = ["false"] //or "true"
  }
}
```

Maybe merged PR #471 which handled a similar error for `tags` can point to how to fix this?

PR #714, opened 2025-01-09
bugfix: query parameters and headers don't work properly with unknown values

ms-henglu:
fixes https://github.com/Azure/terraform-provider-azapi/issues/711

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 FEATURES:
 - **New Ephemeral Resource**: azapi_resource_action
 
+BUG FIXES:
+- Fix a bug that query parameters and headers don't work properly with unknown values
+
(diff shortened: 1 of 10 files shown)


magodo (commented):
Shall we state this is a limitation, instead of silently ignore them and return empty query/header? Ideally, I think we shall split the header/query between different operations, e.g. read_header, create_header, etc.

ms-henglu:
> Shall we state this is a limitation, instead of silently ignore them and return empty query/header? Ideally, I think we shall split the header/query between different operations, e.g. read_header, create_header, etc.

We only convert the headers/query parameters from types.Map to go type in the CRUD functions, these values must be known at that time.

magodo (approved):

Issue #707, opened 2025-01-07
feat: consider adding `sensitive_response_export_values`


matt-FFFFFF:
Several resource properties, or properties returned by POST re4quests can be sensitive.

E.g. list storage account keys, get Log Analytics Workspace shared keys.

## Current behaviour

This results in the output being viewable in the console:

```terraform
# The response is recorded in the state
resource "azapi_resource_action" "log_analytics_workspace_keys" {
  resource_id            = azapi_resource.log_analytics_workspace.id
  type                   = "Microsoft.OperationalInsights/workspaces@2023-09-01"
  action                 = "sharedKeys"
  method                 = "POST"
  response_export_values = ["*"]
}
```

## Proposed behaviour

This results in the output being masked in the console:

```terraform
# The response is recorded in the state
resource "azapi_resource_action" "log_analytics_workspace_keys" {
  resource_id                      = azapi_resource.log_analytics_workspace.id
  type                             = "Microsoft.OperationalInsights/workspaces@2023-09-01"
  action                           = "sharedKeys"
  method                           = "POST"
  sensitive_response_export_values = ["*"]
}
```

This would store the output in a new property, e.g. `sensitive_output`.

WDYT @ms-henglu ?

ms-henglu:
Hi @matt-FFFFFF ,

Thank you for taking time to report this issue!

Sounds good, I'll add this feature in the next release.

Hi @stemaMSFT , do you have any thoughts/concerns about the naming?

stemaMSFT:
Nothing against `sensitive_response_export_values` as it's descriptive and builds on the original function. Should the default be empty @matt-FFFFFF ?

matt-FFFFFF:
> Nothing against `sensitive_response_export_values` as it's descriptive and builds on the original function. Should the default be empty @matt-FFFFFF ?

Yes I think leave it to the caller to discern

PR #718, opened 2025-01-10
`azapi_resource_action` - support `sensitive_response_export_values` and `sensitive_output` fields

ms-henglu:
fixes https://github.com/Azure/terraform-provider-azapi/issues/707


matt-FFFFFF:
Hi @ms-henglu do you think this could be in place for azapi_resource as well?

internal/docstrings/output.go
```diff
@@ -22,3 +22,24 @@ const (
...
 	return addBackquotes(strings.ReplaceAll(outputStr, "RESOURCE", res))
 }
+
+const (
+	sensitiveOutputStr = `The output HCL object containing the properties specified in %ssensitive_response_export_values%s. Here are some examples to use the values.
+
+	%s%s%sterraform
+	// it will output "registry1.azurecr.io"
+	output "login_server" {
+		value = RESOURCE.example.output.properties.loginServer
```
matt-FFFFFF, line 32: Should say .sensitive_output. I think

internal/docstrings/output.go
```diff
@@ -22,3 +22,24 @@ const (
...
+	}
+
+	// it will output "disabled"
+	output "quarantine_policy" {
+		value = RESOURCE.example.output.properties.policies.quarantinePolicy.status
```
matt-FFFFFF, line 37: Again should say sensitive_output I think

internal/docstrings/output.go
ms-henglu: Good catch. I'll fix it.

ms-henglu:
Hi @matt-FFFFFF ,

Thanks for the code review!

I think it's not necessary to add it in the `azapi_resource`, because `azapi_resource` only calls `GET` to the resource, and `GET` response should not contain sensitive values.

magodo (approved):

PR #729, opened 2025-01-16
fix(retry): add WithMaxElapsedTime to retry options and refactor to use functional options

matt-FFFFFF:
@jaredfholgate discovered an issue with long resource operations and retry.

The package we use has a [default max elapsed time of 15m](https://github.com/cenkalti/backoff/blob/720b78985a65c0452fd37bb155c7cac4157a7c45/exponential.go#L83).

This causes an issue when the resource's `timeout` values were set to longer than 15m.

This PR adds the `backoff.WithMaxElapsedTime()` setting into the retry configuration. It sets this to the resource's timeout value. Since we use the context aware retry function in the backoff package this value just needs to be equal to or higher than the timeout value. Once the context deadline is exceeded then the retry package will exit anyway (we already have unit tests for this).

I have also refactored to remove the unnecessary `NewRetryableErrors()` function from the clients package. This was inflexible and using the functional options variadic pattern provided by the backoff package is much cleaner.

To support the conversion of `[]string` to `[]regexp` I have added a small function.

WDYT @ms-henglu ?

### DIFF
--- a/internal/clients/data_plane_client_test.go
+++ b/internal/clients/data_plane_client_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/Azure/terraform-provider-azapi/internal/clients"
 	"github.com/Azure/terraform-provider-azapi/internal/services/parse"
+	"github.com/cenkalti/backoff/v4"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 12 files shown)


ms-henglu (approved):
LGTM

ms-henglu:
Thanks @matt-FFFFFF for this PR!

PR #730, opened 2025-01-16
feat: ephemeral resource sensitive output

matt-FFFFFF:
Given that ephemeral resources are likely to be sensitive, do we need to add the sensitive outputs from `azapi_resource_action`?

WDYT @ms-henglu


ms-henglu:
Hi @matt-FFFFFF ,

Thank you for taking time to make this PR. 

I think it's not necessary to add a sensitive output, because ephemeral resource won't be stored in terraform state and also not showed in terraform plan.

matt-FFFFFF:
No problem. I'll close. Thank you

PR #732, opened 2025-01-20
Add example for Microsoft.DBforPostgreSQL_flexibleServers_configurations@2022-12-01

neil-yechenwei:
Add support for https://github.com/ms-henglu/tools/blob/main/tools/azurerm-issue-comment-searcher/examples/2331080928/comment.md

image

### DIFF
--- a/examples/Microsoft.DBforPostgreSQL_flexibleServers_configurations@2022-12-01/main.tf
+++ b/examples/Microsoft.DBforPostgreSQL_flexibleServers_configurations@2022-12-01/main.tf
@@ -3,14 +3,6 @@ terraform {
     azapi = {
       source = "Azure/azapi"
     }
-    azurerm = {
-      source = "hashicorp/azurerm"
-    }
-  }
-}
-
-provider "azurerm" {
-  features {
   }
 }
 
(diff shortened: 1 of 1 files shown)


examples/Microsoft.DBforPostgreSQL_flexibleServers_configurations@2022-12-01/main.tf
```diff
@@ -0,0 +1,96 @@
+terraform {
+  required_providers {
+    azapi = {
+      source = "Azure/azapi"
+    }
+    azurerm = {
```
ms-henglu, line 6: Please remove the azurerm definition and its provider block if it's not used.

examples/Microsoft.DBforPostgreSQL_flexibleServers_configurations@2022-12-01/main.tf
```diff
@@ -0,0 +1,96 @@
...
+      source = "hashicorp/azurerm"
+    }
+  }
+}
+
+provider "azurerm" {
```
ms-henglu, line 12: Same here.

ms-henglu (changes requested):
Thanks for the PR, it mostly LGTM, just one minor suggestion, please check.

examples/Microsoft.DBforPostgreSQL_flexibleServers_configurations@2022-12-01/main.tf
neil-yechenwei: Updated

Issue #691, opened 2024-12-10
`data.azapi_resource` does not handle `429 Too Many Requests`


tiwood:
We're using the provider to get a list of list of available service tags for our Azure Firewall automation (via `https://management.azure.com/subscriptions/.../providers/Microsoft.Network/locations/westeurope/serviceTags`. This API is rate limited, which is currently not handled by the provider.

```shell
  │ --------------------------------------------------------------------------------
  │ RESPONSE 429: 429 Too Many Requests
  │ ERROR CODE: RetryableError
  │ --------------------------------------------------------------------------------
  │ {
  │   "error": {
  │     "code": "RetryableError",
  │     "message": "A retryable error occurred.",
  │     "details": [
  │       {
  │         "code": "RetryableErrorDueToTooManyCallsToThisOperation",
  │         "message": "Operation GetServiceTagsOperation exceeded throttling limit of 5 calls within last 5 minutes. The number of calls exceeds Microsoft.Network throttling limit. The call can be retried in 76 seconds."
  │       }
  │     ]
  │   }
  │ }
  │ --------------------------------------------------------------------------------
```

I would expect that the provider handles these automatically, honoring the related `rety-after` headers.

tiwood:
I've solved it via the `retry` argument:

```terraform
  retry = {
    error_message_regex = ["A retryable error occurred."]
  }
```

If you ask me, I think retryable errors should be retried without any user configuration.

matt-FFFFFF:
429 errors are handled by the Azure go sdk

https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/policy_retry.go

@hqhqhqhqhqhqhqhqhqhqhq I think the default will retry 3 times on 429 errors.

Perhaps we should make sure we are using this policy?

matt-FFFFFF:
In pr #745 I expose the underlying GO SDK's retry count.

This could be a solution to this?

PR #735, opened 2025-01-22
feat: extract retry config from resources and added default retry config

hqhqhqhqhqhqhqhqhqhqhq:
This pr extract the repetitive code in setting up client with retry and added default retry configs when custom retry logics not provided.
Also solves https://github.com/Azure/terraform-provider-azapi/issues/691


internal/services/azapi_resource.go
```diff
@@ -727,21 +715,10 @@ func (r *AzapiResource) CreateUpdate(ctx context.Context, requestPlan tfsdk.Plan
...
-		[]int{404},
-		[]func(d interface{}) bool{
-			func(d interface{}) bool {
-				return d == nil
-			},
-		},
-	)
+
+	// Create a new retry client to handle specific case of transient 404 or empty body after resource creation or other retryable errors
+	clientGetAfterPut := r.ProviderData.ResourceClient.ConfigureClientWithCustomRetry(ctx, plan.Retry)
```
matt-FFFFFF, line 720: The retry for the get after put is a special case and does not use the configuration in the resource's retry attribute. Removing the above code will cause a regression for resources deployed at MG scope, as this relies on the response check function:

```
			func(d interface{}) bool {
				return d == nil
			},
```

ms-henglu:
Close this PR in favor of https://github.com/Azure/terraform-provider-azapi/pull/745

PR #745, opened 2025-01-31
feat: consideration for retry v2

matt-FFFFFF:
This PR builds on and enhances the excellent work of @hqhqhqhqhqhqhqhqhqhqhq in #735.

Apologies for the diff, I had to update some framework dependencies and vendor them.

- It fixes the issue where the max timeout for read after create could only be set via environment
- Exposes the Azure Go SDK's max retry value in a provider configuration. This is the optimum way of retrying 429 type requests with a `Retry-After` header.
- Adds more detailed documentation in the form of a guide

## Default retry

By default the provider will retry for on a 429 error. This is a catch-all in case the go SDK's max retries have been exceeded. I have removed the nil response retry from here as this triggers on 403 events as well.

## Default retry for read after create requests

The default behaviour is to retry on 404 and 403 events. The timeout is the same as that defined in the `timeouts {}` block. This makes sense as, if we have just created the resource we should ignore any 403/404 events as these are highly likely to be due to eventual consistency issues.

## Clarification of retry on 429/408, etc

In the first instance handled by the Go SDK's Retry policy. This is preferable I believe as this will honour the `Retry-After` response header.

## Chores

To use the new framework types of Float64 and Int64 I had to bump the following dependencies:

- bumped github.com/hashicorp/terraform-plugin-sdk/v2
- bumped github.com/hashicorp/terraform-plugin-go
- vendoring updates

FYI @stemaMSFT @ms-henglu


docs/data-sources/resource.md
```diff
@@ -136,14 +136,17 @@ To learn more about JMESPath, visit [JMESPath](https://jmespath.org/).
...
-- `error_message_regex` (List of String) A list of regular expressions to match against error messages. If any of the regular expressions match, the error is considered retryable.
+- `error_message_regex` (List of String)
 
 Optional:
 
 - `interval_seconds` (Number) The base number of seconds to wait between retries. Default is `10`.
 - `max_interval_seconds` (Number) The maximum number of seconds to wait between retries. Default is `180`.
 - `multiplier` (Number) The multiplier to apply to the interval between retries. Default is `1.5`.
 - `randomization_factor` (Number) The randomization factor to apply to the interval between retries. The formula for the randomized interval is: `RetryInterval * (random value in range [1 - RandomizationFactor, 1 + RandomizationFactor])`. Therefore set to zero `0.0` for no randomization. Default is `0.5`.
+- `response_is_nil` (Boolean) If set to `true`, the retry will be triggered when the response is `nil`. Default is `true`.
```
jaredfholgate, line 147: Do these combine with or override the `error_message_regex`? The `error_message_regex` is marked as a required input still?

matt-FFFFFF: Good shout. Will change

docs/data-sources/resource.md
```diff
@@ -136,14 +136,17 @@ To learn more about JMESPath, visit [JMESPath](https://jmespath.org/).
 
 Required:
 
-- `error_message_regex` (List of String) A list of regular expressions to match against error messages. If any of the regular expressions match, the error is considered retryable.
+- `error_message_regex` (List of String)
```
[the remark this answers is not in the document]
matt-FFFFFF: No... will revert thanks

internal/provider/provider.go
```diff
@@ -644,11 +650,15 @@ func (p Provider) Configure(ctx context.Context, request provider.ConfigureReque
 		response.Diagnostics.AddError("Failed to obtain a credential.", err.Error())
 		return
 	}
-
+	maxGoSdkRetryAttempts := int32(3)
```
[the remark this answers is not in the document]
matt-FFFFFF: This is the underlying go SDKs retry mechanic. The workings of it are here: <https://github.com/Azure/azure-sdk-for-go/blob/90c29ccd6952363d37f7f2c2aa2e8d329ee59015/sdk/azcore/policy/policy.go#L92>

It's designed to handle 429s, etc and then retries according to the `Retry-After` header. Because it understands the response header it's better at handling this than an exponential backoff.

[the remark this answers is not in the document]
jaredfholgate: Ok, makes sense. Re-reading the new doc page, it does explain this.

jaredfholgate:
The one thing I am still unsure about is how these 4 work together:

- `error_message_regex`
- `response_is_nil`
- `status_forbidden`
- `status_not_found`

Are they combined and de-duplicated?

If so, it would be good to call that our in the docs.

docs/guides/retry-configuration.md
```diff
@@ -0,0 +1,60 @@
...
+There are two types of retry configurations available in the AzAPI provider:
+
+1. Provider retry configuration
+2. Resource-specific retry configuration
+
+### Provider retry configuration
+
+The provider retry configuration is a global configuration that applies to all resources managed by the provider. You can configure the provider retry behavior by setting the following provider values:
+
+- `go_sdk_maximum_retry_attempts`
```
cwe1ss, line 23: The `go_sdk` prefix seems like an unnecessary / confusing implementation detail to me. As a user I might not even know that providers are implemented in go. I think it would be better to use just `maximum_retry_attempts`.

matt-FFFFFF: It was to differentiate it from the resource specific retry. But I agree a better name is probably available. Will have a think!

matt-FFFFFF:
> @matt-FFFFFF The one thing I am still unsure about is how these 4 work together:
> 
> * `error_message_regex`
> * `response_is_nil`
> * `status_forbidden`
> * `status_not_found`
> 
> Are they combined and de-duplicated?
> 
> If so, it would be good to call that out in the docs.

Added into guide

docs/guides/retry-configuration.md
matt-FFFFFF: renamed... thoughts?

cwe1ss:
Just wanted to say that this will be a great addition to the provider. Thank you for your efforts on this!

cwe1ss:
Would it be possible to add a provider-level flag / environment variable for retries on 403/404 as part of this MR? We use AzAPI for our subscription-vending and as part of that, our service principal often creates a resource group, assigns itself additional permissions, and then creates the resources within the RG. And that often temporarily fails with a 403 because the Azure-internal replication isn't fully done.

A provider-level flag/environment variable would ensure that we don't have to set these retry-flags on each and every resource.

jaredfholgate:
> Would it be possible to add a provider-level flag / environment variable for retries on 403/404 as part of this PR? We use AzAPI for our subscription-vending and as part of that, our service principal often creates a resource group, assigns itself additional permissions, and then creates the resources within the RG. And that often temporarily fails with a 403 because the Azure-internal replication isn't fully done.
> 
> A provider-level flag/environment variable would ensure that we don't have to set these retry-flags on each and every resource.

I don't think we'll implement this at the provider level, at least not in this PR. For you use case, would a locals block make it easier for you. Here is an example where we do that (define once, use many): https://github.com/Azure/alz-terraform-accelerator/blob/800b976a99d3b3330f02443c49ad6dc11d01cf99/templates/platform_landing_zone/modules/management_groups/locals.tf#L16

docs/guides/retry-configuration.md
```diff
@@ -0,0 +1,59 @@
+---
+layout: "azapi"
+page_title: "Feature: Retry Configuration"
+description: |-
+  This guide will describe how to configure the retry behavior in the AzAPI provider. The retry configuration allows you to control how the provider handles failed requests and if it should retry them.
+---
+
+## Why retry?
```
ms-henglu, line 8: Thanks for the doc! Actually we have a doc for the same purpose, would you please combine the two docs?
refs: https://github.com/Azure/terraform-provider-azapi/blob/main/docs/guides/feature_customized_retry.md

internal/retry/retryable_errors.go
```diff
@@ -18,38 +17,49 @@ import (
 var (
 	_ basetypes.ObjectTypable  = RetryType{}
 	_ basetypes.ObjectValuable = RetryValue{}
+
+	defaultRetryableStatusCodes                = []int{429}
```
ms-henglu, line 21: I think 429 retry is already covered by the azure SDK?

internal/retry/retryable_errors.go
```diff
@@ -18,38 +17,49 @@ import (
...
+	defaultRetryableReadAfterCreateStatusCodes = []int{404, 403}
```
ms-henglu, line 22: I think 403 retry in read-after-create only applied to some of the resources, it might not be a common behavior. So maybe use 404 as the default retry status codes for read-after-create?

ms-henglu (changes requested):
Thanks Matt for this PR! It mostly LGTM, just some minor suggestions, please take a look. Once they're addressed, I'll approve and merge this PR.

Thanks again.

internal/clients/client_configure_retry.go
```diff
@@ -0,0 +1,52 @@
...
+	backOff := backoff.NewExponentialBackOff(
+		backoff.WithInitialInterval(retry.DefaultIntervalSeconds*time.Second),
+		backoff.WithMaxInterval(retry.DefaultMaxIntervalSeconds*time.Second),
+		backoff.WithMaxElapsedTime(maxElapsed),
+	)
+	errRegExps := []regexp.Regexp{}
+	statusCodes := rtry.GetDefaultRetryableStatusCodes()
+	// Add default read after create values for the default retry configuration.
+	if useReadAfterCreateValues {
+		statusCodes = rtry.GetDefaultRetryableReadAfterCreateStatusCodes()
```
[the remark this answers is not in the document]
matt-FFFFFF: This was not the intent here. The get after put retry status codes were intended to be different. I could add 429 to the default codes returned for this operation, however I don't think it's necessary based on your other comments.

internal/retry/retryable_errors.go
matt-FFFFFF: This acted as a safety net, but I agree probably not necessary given the new provider level configuration.

internal/retry/retryable_errors.go
matt-FFFFFF: My logic here was thus: Since we have successfully created the resource (received a HTTP 200), then we should never receive a 403 or 403 when we try to GET the resource. Hence this suggestion.

Given that now there is no way to specifically alter the behaviour of the this request, then I think this is a sensible decision. WDYT?

matt-FFFFFF:
@ms-henglu I have addressed the comments. Please would you review my responses and the recent changes? Thank you

ms-henglu (approved):

ms-henglu:
Thanks @matt-FFFFFF for this PR. LGTM

PR #756, opened 2025-02-11
Add example for Microsoft.OperationalInsights_workspaces_tables@2022-10-01

neil-yechenwei:
This PR is to support https://github.com/ms-henglu/tools/blob/main/tools/azurerm-issue-comment-searcher/examples/1876440882/comment.md.

image


ms-henglu:
Thanks for the PR, I'll close this one because we have similar examples for this resource.

PR #764, opened 2025-02-12
Add example for Microsoft.Storage/storageAccounts/privateEndpointConnections@2021-09-01

neil-yechenwei:
This PR is to support https://github.com/ms-henglu/tools/blob/main/tools/azurerm-issue-comment-searcher/examples/1706220254/comment.md.

image


ms-henglu:
Thanks for the PR, but I'll close this, because the original example can demonstrate how to use storage account with a private endpoint.

PR #765, opened 2025-02-12
Add example for Microsoft.Web_connections@2016-06-01

neil-yechenwei:
This PR is to support https://github.com/ms-henglu/tools/blob/main/tools/azurerm-issue-comment-searcher/examples/1702491304/comment.md.

image

### DIFF
--- a/examples/Microsoft.Web_connections@2016-06-01/main.tf
+++ b/examples/Microsoft.Web_connections@2016-06-01/main.tf
@@ -12,7 +12,6 @@ provider "azapi" {
 
 provider "azurerm" {
   features {
-
   }
 }
 
@@ -83,6 +82,8 @@ resource "azapi_resource" "namespaces" {
 data "azurerm_managed_api" "test" {
   name     = "servicebus"
   location = var.location
+
+  depends_on = [azapi_resource.workflows, azapi_resource.namespaces]
 }
 
 resource "azapi_resource" "connection" {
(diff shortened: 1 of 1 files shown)


examples/Microsoft.Web_connections@2016-06-01/main.tf
```diff
@@ -0,0 +1,103 @@
...
+  schema_validation_enabled = false
+  response_export_values    = ["*"]
+}
+
+data "azurerm_managed_api" "test" {
+  name     = "servicebus"
+  location = var.location
+}
+
+resource "azapi_resource" "connections" {
```
ms-henglu, line 88: Please change the label to `connection`

ms-henglu (changes requested):

examples/Microsoft.Web_connections@2016-06-01/main.tf
neil-yechenwei: Updated

neil-yechenwei:
@ms-henglu , please take another review. thx

examples/Microsoft.Web_connections@2016-06-01/main.tf
```diff
@@ -0,0 +1,103 @@
...
+  schema_validation_enabled = false
+  response_export_values    = ["*"]
+}
+
+data "azurerm_managed_api" "test" {
+  name     = "servicebus"
+  location = var.location
+}
+
+resource "azapi_resource" "connection" {
```
ms-henglu, line 88: It seems that this resource doesn't require the workflow and service bus namespace defined above?

neil-yechenwei: The connection resource requires them.

ms-henglu (approved):

PR #772, opened 2025-02-18
Add example for Microsoft.OperationalInsights/workspaces/tables@2022-10-01

neil-yechenwei:
This PR is to support https://github.com/ms-henglu/tools/blob/main/tools/azurerm-issue-comment-searcher/examples/1523362914/comment.md.

image


ms-henglu:
Close this PR as the original example demonstrates how to use the resource type.

PR #786, opened 2025-02-21
feat: add skip external request

matt-FFFFFF:
## Skip external API request

The `skip_on` struct field tag is used to skip the external API call when only attributes that affect the internal state are modified, e.g. retry configuration. The `skip_on` struct field tag is a comma-separated list of operations that must be met in order to skip the field.

The provider will compare the state with the plan, and check for changes. If the only fields to be modified are those with the `skip_on` struct field tag set to the supplied operation, e.g. `update`, the provider will skip the external API call.

Only update is currently supported, as I think that's the only time this makes sense.

Separated from #745 as requested my @ms-henglu


ms-henglu (approved):
Thanks for the PR, LGTM.

PR #811, opened 2025-03-13
replace terraform-plugin-sdk/v2 with terraform-plugin-testing

ms-henglu:
This PR removes terraform-plugin-sdk/v2 in testing and uses terraform-plugin-testing


ms-henglu:
The e2e failure is caused by intermittent service error. Not related to this change.

magodo (approved):

PR #814, opened 2025-03-18
Add example for Billing Profile

neil-yechenwei:
This PR is to add example for [Microsoft.Billing_billingAccounts_billingProfiles@2024-04-01](https://github.com/Azure/azure-rest-api-specs/blob/115eea12f7c785b58ff6499d1ad509152bb63c60/specification/billing/resource-manager/Microsoft.Billing/stable/2024-04-01/billingProfile.json#L170).


ms-henglu:
Hi @neil-yechenwei , thanks for the PR. Could you also attach a test evidence?

PR #836, opened 2025-04-06
fix(azapi_resource): set context deadline before config retry

matt-FFFFFF:
@ms-henglu

Please see this small fix that was missed on the last update to the retry code.

This fixes the Delete() method of azapi_resource and brings it in line with the other CRUD methods. This fix ensures that the retry client has the correct max elapsed time, rahter than the default of 15 minutes.


ms-henglu (approved):
Thanks for the PR, LGTM.

PR #838, opened 2025-04-08
bugfix: `azapi_resource` cannot be updated correctly after moving from `azurerm` provider

ms-henglu:
After moving resource, `azapi` sets the private data to indicate the resource is moved from `azurerm` provider, but in the read func, this flag is not removed. This will cause the resource could not be updated correctly.

This PR fixes this bug and adds a test case for it.


magodo (approved):

ms-henglu:
cc @lonegunmanb

PR #842, opened 2025-04-14
remove password from example

grayzu:
removed password from example replacing it with a required variable marked as secret


ms-henglu (approved):
Thanks for the PR, LGTM

PR #852, opened 2025-04-25
fix: missing env vars for azure devops and bug with auth config

jaredfholgate:
This PR adds 2 missing environment variable for `oidc_request_token` and  `oidc_request_url` to bring it inline with the azurerm provider and the azurerm backend.

It also adds a new env var for `oidc_azure_service_connection_id`, so that it automatically works with the Azure DevOps Azure CLI task.

Also fixes this error when attempting to use Azure DevOps auth. It tries to create a standard OIDC client that targets GitHub interfaces and fails. The PR adds a check to see if the service connection ID has been supplied and then creates the correct tpye of authenticator.

```
╷
│ Error: Failed to retrieve resource
│ 
│   with module.virtual_network.azapi_resource.vnet,
│   on .terraform/modules/virtual_network/main.virtual.network.tf line 1, in resource "azapi_resource" "vnet":
│    1: resource "azapi_resource" "vnet" {
│ 
│ reading Resource: (ResourceId
│ "/subscriptions/b857908d-3f5c-4477-91c1-0fbd08df4e88/resourceGroups/rg-dema-env-dev-uksouth-001/providers/Microsoft.Network/virtualNetworks/vnet-demo-dev-uksouth-001"
│ / Api Version "2023-11-01"): ChainedTokenCredential: failed to acquire a
│ token.
│ Attempted credentials:
│ 	ClientAssertionCredential: getAssertion: received HTTP status 405 with
│ response: {"count":1,"value":{"Message":"The requested resource does not
│ support http method 'GET'."}}
╵

```


internal/provider/provider.go
```diff
@@ -262,12 +262,12 @@ func (p Provider) Schema(ctx context.Context, request provider.SchemaRequest, re
...
 			"oidc_request_token": schema.StringAttribute{
 				Optional:            true,
-				MarkdownDescription: "The bearer token for the request to the OIDC provider. This can also be sourced from the `ARM_OIDC_REQUEST_TOKEN` or `ACTIONS_ID_TOKEN_REQUEST_TOKEN` Environment Variables.",
+				MarkdownDescription: "The bearer token for the request to the OIDC provider. This can also be sourced from the `ARM_OIDC_REQUEST_TOKEN`, `ACTIONS_ID_TOKEN_REQUEST_TOKEN`, or `SYSTEM_ACCESSTOKEN` Environment Variables.",
 			},
 
 			"oidc_request_url": schema.StringAttribute{
 				Optional:            true,
-				MarkdownDescription: "The URL for the OIDC provider from which to request an ID token. This can also be sourced from the `ARM_OIDC_REQUEST_URL` or `ACTIONS_ID_TOKEN_REQUEST_URL` Environment Variables.",
+				MarkdownDescription: "The URL for the OIDC provider from which to request an ID token. This can also be sourced from the `ARM_OIDC_REQUEST_URL`, `ACTIONS_ID_TOKEN_REQUEST_URL`, or `SYSTEM_OIDCREQUESTURI` Environment Variables.",
```
ms-henglu, line 270: Please also update below source code to get the value from `SYSTEM_OIDCREQUESTURI`

https://github.com/Azure/terraform-provider-azapi/pull/852/files#diff-58d6a027753b50994deb7e11e4a99dde423f35844986019bd9cea5e0c94aba22L507

ms-henglu (changes requested):
Thanks @jaredfholgate for the PR. LGTM just one minor suggestion, please check.

internal/provider/provider.go
jaredfholgate: It is hard-coded here, so removed it from this code. I had added it origianlly, but wouldn't be used: https://github.com/Azure/azure-sdk-for-go/blob/078e06aa73fb6a8dc585ad4ce1b371780aef2237/sdk/azidentity/azure_pipelines_credential.go#L22

ms-henglu (approved):

internal/provider/provider.go
ms-henglu: Thanks for the clarification!

Issue #858, opened 2025-04-29
importing resource with replace_triggers_external_values recreates the resource


mysteq:
Is it possible to import a resource with a value set for "replace_triggers_external_values" without it recreating the resource?

ms-henglu:
Hi @mysteq ,

Thank you for taking time to report this issue.

I'll work on adding support for this feature: set values for the meta attributes(replace_triggers_external_values, response_export_values..) when import a resource.


A naive solution is adding them in the query parameter. For example, 

```
terraform import azapi_resource.resourceGroup /subscriptions/xxx/resourceGroups/myRG?api-version=2025-01-01& replace_triggers_external_values=["foo"]&response_export_values={"foo": "properties.Foo"}
``` 

What do you think?

mysteq:
Hi @ms-henglu, 

For my use case, this seems like a workable solution that would solve my needs.

jaredfholgate:
@ms-henglu This also impacts `moved` when moving from `azurerm` to `azapi`.

jaredfholgate:
@ms-henglu I think the issue is the code here: https://github.com/Azure/terraform-provider-azapi/blob/ad72a0bca9f29818bc18c229c61d8d6671fd6d89/internal/services/azapi_resource.go#L211

When going from `null` to having a value, it should not cause a replace? @matt-FFFFFF can explain this better than me though.

jaredfholgate:
A couple of reproductions:

```
moved {
  from = module.hub_and_spoke_vnet[0].module.bastion_host["primary"].azurerm_bastion_host.this
  to = module.hub_and_spoke_vnet[0].module.bastion_host["primary"].azapi_resource.bastion[0]
}
```

Results in:

```
# module.hub_and_spoke_vnet[0].module.bastion_host["primary"].azapi_resource.bastion[0] must be replaced
  # (moved from module.hub_and_spoke_vnet[0].module.bastion_host["primary"].azurerm_bastion_host.this)
-/+ resource "azapi_resource" "bastion" {
      ~ body                             = {
          ~ properties = {
              - dnsName                  = "bst-be97745c-695c-498f-bc53-97629844aa01.bastion.azure.com"
              + enableFileCopy           = false
              + enableIpConnect          = false
              + enableKerberos           = false
              + enablePrivateOnlyBastion = false
              + enableSessionRecording   = false
              + enableShareableLink      = false
              + enableTunneling          = false
              ~ ipConfigurations         = [
[...]

PR #866, opened 2025-05-12
fix(planmodifier): requires replace external values should not replace if state value is null

matt-FFFFFF:
fixes importing resource with replace_triggers_external_values recreates the resource #858


matt-FFFFFF:
@ms-henglu please could you review my PR that I hope will fix the issue above?

ms-henglu (approved):
Thanks for the PR! LGTM

PR #874, opened 2025-05-15
Moving all the usage examples under a reference directory

stemaMSFT:


ms-henglu:
Thanks @stemaMSFT for this PR. As we discussed offline, the hashicorp registry doesn't support this child folder. I'll close this PR.

PR #916, opened 2025-07-01
No recreate for `azapi_resource_action` resource on type change

bloewes:
The `azapi_resource_action` resource will force re-creation if the type field changes. This means that simply updating the api-version (for example, from 2022-05-01 to 2023-01-01) can cause the resource to be destroyed and recreated, which is problematic—especially if `when = "destroy"` is set. The resource should still be sufficiently re-created if any of the other marked arguments change.

### DIFF
--- a/internal/services/azapi_resource_action_resource.go
+++ b/internal/services/azapi_resource_action_resource.go
@@ -95,9 +95,6 @@ func (r *ActionResource) Schema(ctx context.Context, request resource.SchemaRequ
 				Validators: []validator.String{
 					myvalidator.StringIsResourceType(),
 				},
-				PlanModifiers: []planmodifier.String{
-					stringplanmodifier.RequiresReplace(),
-				},
 				MarkdownDescription: docstrings.Type(),
 			},
 


bloewes:
For the case `when="destroy"`: since the action hasn’t happened yet, this change makes sense and shouldn’t have any side effects. Is it possible to make the re-creation of the resource’s type conditional on the value of `when`?

Otherwise, with many resources and resource instances that we use and frequent API changes, this approach doesn’t seem effective for my our case. I prefer to continue using `ignore_changes = [type]` in that case.

ms-henglu:
Hi @bloewes , I just realized that even the action resource is not recreated, it'll still trigger one http request if `when != destroy`, is it okay for your case?

bloewes:
@ms-henglu For my case, I only care about when = destory, because those would make a DELETE api call which is then not recoverable. But if the case when = destory does not trigger any call, I'm fine.

ms-henglu:
I'll close this PR in favor of https://github.com/Azure/terraform-provider-azapi/pull/1035, hope you don't mind. Let me know if you have any questions. Thanks for the PR again!

PR #918, opened 2025-07-03
[WIP] consideration: general purpose data plane resources

matt-FFFFFF:
I have implemented a general purpose resource type that is capable of authenticating to any Entra-backed endpoint. It is a low level resource, with no schema validation (this is impossible, as we already know).
 
The benefits of this approach are that we can use Terraform to managing any data plane resource/any Entra backed endpoint.
 
The drawbacks are that this requires knowledge of the API that you are calling.
 
For us, in the AVM team, having this ability would open up a lot of opportunities and be a huge benefit. I propose that we could release this with documentation that this is an advanced resource type with warnings, etc.
 
I would welcome your feedback please!
 
Here is a working code sample of the resource in action:

```hcl
resource "azapi_resource" "keyvault" {
  type      = "Microsoft.KeyVault/vaults@2024-11-01"
  parent_id = azapi_resource.rg.id
  name      = "testazapikvwq"
  location  = azapi_resource.rg.location
  body = {
    # Body properties
  }
  response_export_values = [
    "properties.vaultUri",
  ]
}

resource "azapi_general_resource" "secret" {
  url         = "${azapi_resource.keyvault.output.properties.vaultUri}secrets/test"
  api_version = "7.5"
  body = {
    attributes = {
      enabled = true
    }
  }
  sensitive_body = {
    value = "test" # this would be a reference to an ephemeral value
  }
  sensitive_body_version = {
    "value" = "0"
  }
  response_export_values = [
    "attributes.created",
    "id"
  ]
}
```

If you like this I can continue with:
 
- azapi_general_update_resource
- azapi_general_resource_action
- azapi_general data source

Am obviously happy to rename resources, etc. Naming isn't my strong point!

### DIFF
--- a/docs/resources/general_resource.md
+++ b/docs/resources/general_resource.md
@@ -30,6 +30,8 @@ This resource can manage general resources.
 - `delete_query_parameters` (Map of List of String) A mapping of query parameters to be sent with the delete request.
 - `ignore_casing` (Boolean) A dynamic attribute that contains the request body.
 - `ignore_missing_property` (Boolean) Whether ignore not returned properties like credentials in `body` to suppress plan-diff. Defaults to `true`. It's recommend to enable this option when some sensitive properties are not returned in response body, instead of setting them in `lifecycle.ignore_changes` because it will make the sensitive fields unable to update.
+- `ignore_null_property` (Boolean) When set to `true`, the provider will ignore properties whose values are `null` in the `body`.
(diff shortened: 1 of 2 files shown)


matt-FFFFFF:
@simongottschlag thanks for your interest here. I've been in discussions with the engineers and PMs on this topic to find the next way forward.

In terms of your ask, yes this solution should work with your use case.

I think we are all looking for a usable provider for the Azure data plane.

PR #920, opened 2025-07-09
Feature refactor retry

ms-henglu:
This PR refactors the retry logics to use the retry policy provided by azure-sdk-for-go. So the retry configuration can be passed to client's functions as an option. 

This PR also marks `multiplier` and `randomization_factor` as deprecated and will be removed in the next major release. I made this change because the retry policy doesn't provide these options and I also think these options are too complex for users.


ms-henglu:
Hi @matt-FFFFFF , I refactored the retry part, so the retry config can be passed as a request option. In the future, I'll provide an interface to allow customization for some resources(add delete support for key vault keys). So this change make it easier to design the resource customization interface. Please help review, thanks!

magodo (approved):

internal/clients/options.go
```diff
@@ -1,10 +1,157 @@
...
+					for _, re := range rtry.GetErrorMessagesRegex() {
+						if re.MatchString(responseErr.Error()) {
+							log.Printf("[DEBUG] Retrying request due to error: %s matches regex %s", responseErr.Error(), re.String())
+							return true
+						}
+					}
+				}
+			}
+			return false
+		},
```
matt-FFFFFF, line 141: This can be simplified/improved for readability I think, this way there is only one loop to check the regexps and we return early if the errorMsg is empty.

Comment also still stands regarding structured logging with tflog.


```suggestion
    ShouldRetry: func(resp *http.Response, err error) bool {
      // Get the error message to check against regex patterns,
      // If use the err.Error() string first, else get the response error from the HTTP response.
      var errorMsg string
      if err != nil {
          errorMsg = err.Error()
      } else {
        responseErr := runtime.NewResponseError(resp)
        if responseErr != nil {
            errorMsg = responseErr.Error()
        }
      }

      // Check if the error message matches any of the retryable error regexps
      if errorMsg == "" {
        return false
      }

      for _, re := range rtry.GetErrorMessagesRegex() {
        if re.MatchString(errorMsg) {
            log.Printf("[DEBUG] Retrying request due to error: %s matches regex %s", errorMsg, re.String())
            return true
        }
      }

      return false
    },
```

internal/clients/options.go
```diff
@@ -1,10 +1,157 @@
...
+	if len(opts) == 0 {
+		return nil
+	}
+
+	statusCodes := make([]int, 0)
+	for _, opt := range opts {
+		if opt == nil {
+			continue
+		}
+		statusCodes = append(statusCodes, opt.StatusCodes...)
```
matt-FFFFFF, line 45: I think we could avoid duplicate values here?

```go
for _, code := range opt.StatusCodes {
  if slices.Contains(statusCodes, code) {
    append(statusCodes, code)
  }
}
```

internal/clients/options.go
```diff
@@ -1,10 +1,157 @@
...
+		for _, opt := range opts {
+			if opt == nil {
+				continue
+			}
+			if opt.ShouldRetry != nil {
+				if opt.ShouldRetry(resp, err) {
+					return true
+				}
+			}
+		}
```
matt-FFFFFF, line 82: We could keep the happy path more left aligned here

```suggestion
		for _, opt := range opts {
			if opt == nil || opt.ShouldRetry == nil {
				continue
			}

      if opt.ShouldRetry(resp, err) {
				return true
			}
		}
```

internal/clients/options.go
```diff
@@ -1,10 +1,157 @@
...
+func NewRetryOptionsForReadAfterCreate(maxRetries int32) *policy.RetryOptions {
+	statusCodes := make([]int, 0)
+	statusCodes = append(statusCodes, DefaultRetryableStatusCodes...)
+	// Add default read after create values for the default retry configuration.
+	statusCodes = append(statusCodes, DefaultRetryableReadAfterCreateStatusCodes...)
+	return &policy.RetryOptions{
+		MaxRetries:  maxRetries,
+		StatusCodes: statusCodes,
+	}
+}
```
matt-FFFFFF, line 105: The inclusion of a fixed number of retries here will break things for us. This is a regression to a previous behavior when we had a low number of maximum retries for GET after PUT operations.

Implementing this as-is will break Azure Landing Zone deployments, because it can take the Management Groups API up to 20 minutes for a GET after PUT operation to succeed.

We must essentially have no cap on teh maximum number of retried and instead use the resource create timeout value (context deadline), this was a MUCH better approach.

```suggestion
// NewRetryOptionsForReadAfterCreate creates a RetryOptions for read-after-create operations.
func NewRetryOptionsForReadAfterCreate() *policy.RetryOptions {
	statusCodes := make([]int, 0)
	statusCodes = append(statusCodes, DefaultRetryableStatusCodes...)
	// Add default read after create values for the default retry configuration.
	statusCodes = append(statusCodes, DefaultRetryableReadAfterCreateStatusCodes...)
	return &policy.RetryOptions{
		MaxRetries:  math.MaxInt16,
		StatusCodes: statusCodes,
	}
}
```

matt-FFFFFF (commented):
Hi @ms-henglu

FIrst of all great idea to use the Go SDK's built-in ShouldRetry func. Fully supportive of this!

However, there is one critical issue in this PR, the remaining comments are readability/simplification based but do not affect functionality.

The critical issue is this function: `NewRetryOptionsForReadAfterCreate()`

We should not impose a maximum number of retries here (default 3), this is a regression and will break Azure Landing Zones deployments. At the moment we retry GET after PUT operations up to the resource create timeout. This change breaks that behaviour and will cause issues with the policy API at management group scope. This API can take 20 mins before returnign a success for a GET operation. The caller will not know how many retries this is, but it will be a lot!

The current bahavior of retrying up to the context deadline is what is needed here. Having a seperate value for this retry vs. the other retry is confusing. As far as the caller knows, they just want to retry up to the timeout.

FInally, you mentioned future special cases for KeyVault keys/secrets... please can I make an impassioned plea to not have "magic" resources in this provider. It does not solve the real issue of being able to manage data plane resources with AzAPI and I think goes against the ethos of this work.

For example, I also need to be able to manage AppConfiguration key/values with AzAPI. Currently this is not possible as we don't have a generic resource to do so. AzureRM does support these data plane resources, but misses key parts of the data plane API.

Therefore, as a terraform user on Azue, I am stuck.

ms-henglu:
Hi @matt-FFFFFF ,

Thank you so much for the code review.

> The critical issue is this function: NewRetryOptionsForReadAfterCreate()

Yes, you're correct. The retry option for read after create is the default provider behavior, it should not be override by user's specified max retries. I'll fix it.

Thanks again!

internal/clients/options.go
ms-henglu: Got it, I'll fix it.

ms-henglu:
Hi @matt-FFFFFF , I updated the PR and resolved most of your code review comments. 

But I'm hesitated to use `tflog` because it requires to pass `context` to the function that builds retry option. The `context` is only used for log purpose, I think I'll skip that. 

Thanks!

matt-FFFFFF:
> Hi @matt-FFFFFF , I updated the PR and resolved most of your code review comments.
> 
> But I'm hesitated to use `tflog` because it requires to pass `context` to the function that builds retry option. The `context` is only used for log purpose, I think I'll skip that.
> 
> Thanks!

Thank you! This now mimics the existing functionality in 2.5!

LGTM :shipit:

PR #924, opened 2025-07-16
Use `aztfauth` for the (shared) auth logic

magodo:
This PR uses the `aztfauth` package provided by https://github.com/magodo/entrauth to build the credential. This is to make the auth logic shared among different providers (e.g. azapi, ado, etc.).

The logic is almost the same as before, except the following:
- The client id and token file path will do equality check if `UseAKSWorkloadIdentity==true`, between the specified value and an environment variable (if existed). This is now changed to make those environment variables as a lower priority source of these values.
- The certificate file path was assumed to be binary format only. Now it can also be base64 encoded.
- If the current "active" credential is any of OIDC token file based, OIDC Github, the previous implementation will fallback to the next credential if the active one failed to get the assertion (e.g. the token file failed to read, or the github assertion retrieve failed). This is not the case now, in which case the chained token will just stop and error in case failed to get assertion. Ad-hoc supporting this fallback behavior for only these two credentials seems in-consistent among the other credentials. Instead, not falling back even if specified `&azidentity.ChainedTokenCredentialOptions{RetrySources: true}` seems to be an issue of the `azidentity` https://github.com/Azure/azure-sdk-for-go/issues/24935


matt-FFFFFF:
Hi @magodo do you think `entrauth` should be in the Microsoft GitHub org?

ms-henglu (approved):
Thanks for the PR. LGTM

Issue #942, opened 2025-08-01
Error: The terraform-provider-azapi_v2.5.0 plugin crashed!


mmucmp:
Hi,

Been deploying using HCP for months. No changes in the provider version but when we deploy from HCP we now see "Error: The terraform-provider-azapi_v2.5.0 plugin crashed!". Full logs here:

`╷
│ Error: Plugin did not respond
│ 
│ The plugin encountered an error, and failed to respond to the
│ plugin6.(*GRPCProvider).ReadResource call. The plugin logs may contain more
│ details.
╵
╷
│ Error: Plugin did not respond
│ 
│ The plugin encountered an error, and failed to respond to the
│ plugin6.(*GRPCProvider).ReadResource call. The plugin logs may contain more
│ details.
╵

Stack trace from the terraform-provider-azapi_v2.5.0 plugin:

panic: interface conversion: interface {} is float64, not string

goroutine 102 [running]:
github.com/Azure/terraform-provider-azapi/internal/azure/identity.FlattenIdentity({0xd33a80?, 0xc00090e8a0})
	github.com/Azure/terraform-provider-azapi/internal/azure/identity/identity.go:100 +0x46c
github.com/Azure/terraform-provider-azapi/internal/services.(*AzapiResource).Read(0xc00005a4f0, {0x11fb6a58, 0xc000c0e120}, {{{{0x11fbc8f0, 0xc000eaf380}, {0xd36240, 0xc000eae510}}, {0x11fbfe70, 0xc00035cd20}}, 0xc00005a640, ...}, ...)
	github.com/Azure/terraform-provider-azapi/internal/services/azapi_resource.go:929 +0x1655
github.com/hashicorp/terraform-plugin-framework/internal/fwserver.(*Server).ReadResource(0xc0001ae908, {0x11fb6a58, 0xc000c0e120}, 0xc000c0e180, 0xc00056d5d8)
	github.com/hashicorp/terraform-plugin-framework@v1.14.1/internal/fwserver/server_readresource.go:119 +0x84e
github.com/hashicorp/terraform-plugin-framework/internal/proto6server.(*Server).ReadResource(0xc0001ae908, {0x11fb6a58?, 0xc000c0e030?}, 0xc00170e580)
	github.com/hashicorp/terraform-plugin-framework@v1.14.1/internal/proto6server/server_readresource.go:55 +0x38e
[...]

PR #947, opened 2025-08-08
bugfix: provider crash if the `identity.type` returned is not a string

ms-henglu:
fixes https://github.com/Azure/terraform-provider-azapi/issues/942


magodo (approved):

mmucmp:
@ms-henglu when are you releasing 2.6.0?

PR #951, opened 2025-08-11
preflight - fix name field

teowa:
Fixed the name field to align with the number of segments in the resource type field. For example, if the type is "virtualNetworks/subnets", the name will now be "vnetName/subnetName".


ms-henglu (approved):
Thanks for the PR! LGTM

PR #957, opened 2025-08-15
add doc for choosing the resource type

ms-henglu:


mybayern1974:
Thank you for composing this doc, LGTM.

Issue #981, opened 2025-09-22
bug: azapi_client_config data source produces inconsistent results when used with a provider alias


jaredfholgate:
When using a provider alias and `azapi_client_config` it randomly returns different subscription ids on different plans. Sometime it is the correct ones, other times it is the wrong one.

E.g. It results in idempotency issues like this:

```terraform
Changes to Outputs:
  ~ subscription_ids = {
      ~ alias        = "a4225384-b567-4381-9ca4-13f5e2e5ab6c" -> "b857908d-3f5c-4477-91c1-0fbd08df4e88"
      ~ alias_module = "b857908d-3f5c-4477-91c1-0fbd08df4e88" -> "a4225384-b567-4381-9ca4-13f5e2e5ab6c"
        # (1 unchanged attribute hidden)
    }

You can apply this plan to save these new output values to the Terraform state, without changing any real       
infrastructure.
```

We don't see the same issue with the azurrm equivalent data source.

In the repro below we use a sub module, whether that is relevant I am unsure, but this is the scenario we have with our AVM.

Repo code:

Root module:

```terraform
terraform {
  required_providers {
    azapi = {
      source  = "Azure/azapi"
      version = "~> 2.0"
    }
  }
}

provider "azapi" {

}

provider "azapi" {
  alias           = "alias"
  subscription_id = var.alias_subscription_id
}

variable "alias_subscription_id" {
  type        = string
  description = "The subscription ID for the alias provider."
}

data "azapi_client_config" "current" {}

data "azapi_client_config" "alias" {
  provider = azapi.alias
}

module "alias" {
  source = "./module/alias"

  providers = {
    azapi = azapi.alias
  }
}

output "subscription_ids" {
  value = {
    current      = data.azapi_client_config.current.subscription_id
    alias        = data.azapi_client_config.alias.subscription_id
    alias_module = module.alias.subscription_id
  }
}

```

Submodule:

```terraform
terraform {
  required_providers {
    azapi = {
      source  = "Azure/azapi"
      version = "~> 2.0"
    }
  }
}

data "azapi_client_config" "current" {}

output "subscription_id" {
  value = data.azapi_client_config.current.subscription_id
}

```

jaredfholgate:
The provider team started looking into this issue yesterday, so hoping there will be a fix out soon.

Issue #963, opened 2025-08-18
The provider doesn't always use the correct subscription


antoineozenne:
# Description

It seems that the provider `azapi` doesn't always use the correct subscription despite the `subscription_id` parameter.

# To reproduce

## Description

In the following manifests:

* Initialize the `azapi` provider in the root module with the subscription `sub1`.
* Call a child module (20 times) that output a `data.azapi_client_config` declared into it.
* Output the subscription in `data.azapi_client_config` in the root module.
* Output the subscription in `data.azapi_client_config` of each occurrence of the child module. 

```hcl
# main.tf
terraform {
  required_providers {
    azapi = {
      source  = "Azure/azapi"
      version = ">= 2.6.0"
    }
  }

  required_version = ">= 1.12.2"

  backend "azurerm" {
    subscription_id      = "sub1" # Subscription for the state
    resource_group_name  = "rg-tf"
    storage_account_name = "st-tf"
    container_name       = "stct-tf"
    key                  = "tf.tfstate"
    snapshot             = true
    use_azuread_auth     = true
  }
}

provider "azapi" {
  subscription_id = "sub1" # Subscription for resources to deploy
}

data "azapi_client_config" "azapi_client_config" {}

output "azapi_client_config_root" {
  value = data.azapi_client_config.azapi_client_config.subscription_id
}

module "foo" {
  source = "./modules/foo"

  count = 20
}

output "azapi_client_config_childs" {
  value = module.foo.*.azapi_client_config
}
```

```hcl
# ./modules/foo/main.tf
terraform {
  required_providers {
    azapi = {
      source  = "Azure/azapi"
      version = ">= 2.6.0"
    }
  }

  required_version = ">= 1.12.2"
}

data "azapi_client_config" "azapi_client_config" {}

output "azapi_client_config" {
  value = data.azapi_client_config.azapi_client_config
}
```

## Steps

1. Set the current subscription to `sub2` with `az account set --subscription "sub2"`.
2. `terraform init`.
3. `terraform plan`.

## Output

```console
Changes to Outputs:
  + azapi_client_config_childs = [
      + {
          + id                       = "clientConfigs/subscriptionId=sub2;tenantId=XXX"
          + object_id                = "9916287d-445a-4710-874c-1803cc8e1b85"
          + subscription_id          = "sub2"
          + subscription_resource_id = "/subscriptions/sub2"
          + tenant_id                = "XXX"
          + timeouts                 = null
        },
      + {
[...]

PR #983, opened 2025-09-24
bugfix: specified subscription ID should not be overridden by Azure CLI default subscription ID

ms-henglu:
fixes https://github.com/Azure/terraform-provider-azapi/issues/963
fixes https://github.com/Azure/terraform-provider-azapi/issues/981

When use Azure CLI authentication and specify a subscription ID in the `azapi` provider block, the specified subscription ID will be overridden by the subscription ID read from azure cli.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ ENHANCEMENTS:
 - Add documentation for choosing the resource type.
 
 BUG FIXES:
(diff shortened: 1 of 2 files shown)


jaredfholgate (approved):
LGTM

jaredfholgate:
@ms-henglu Do you think we could get this out in a v2.6.2 patch if v2.7.0 is not due in the next week?

PR #1029, opened 2025-12-16
ignoreNotFound for azapi_ressource_action

vincentDcmps:


internal/services/azapi_resource_action_resource.go
matt-FFFFFF: In the case that ignore 404 is set, I think we should add a computed attribute that indicates to the caller whether the action worked, or not. I believe we do the same on the data.azapi_resource

vincentDcmps: yes I can handle that, not included it at beginning because action not represent in the state a dedicated resource

ms-henglu (changes requested):
Thank you for taking time to open this PR!

It mostly LGTM, just some minor comments, please check.

ms-henglu (approved):
Thanks for the PR! LGTM. Please fix the checks.

vincentDcmps:
> Thanks for the PR! LGTM. Please fix the checks.

done

Issue #995, opened 2025-10-14
Feature request: further enhance change detection for arrays


matt-FFFFFF:
I would like to establish whether we can have a feature that will treat lists of objects in a resource body more like merging maps, than lists.

Currently, if the name attribute of the array item is set, it will compare the items in the array based on their names to provide a more accurate change detection. If the name attribute is not set, it will compare the items based on their order in the array.

We have several situations where we want to manage **only certain** items in a list of objects. For example, take the service endpoints feature of a subnet. There are cases where we want to add an additional service endpoint, but do not want to change any existing values.

I propose that we could introduce an advanced change detection feature, something like this:

### Potential solution

Consider a subnet that already had KeyVault service endpoint configured and we wanted to add one more (Web):

```hcl
resource "azapi_resource" "subnet" {
  # other stuff
  body = {
    properties = {
      serviceEndpoints = [
        {
          service = "Microsoft.Web"
        }
      ]
    }
  }

  # The below would tell the provider to treat this list as a map,
  # ensuring that the provided values were present, but not re-ordering/removing
  # others.
  list_unique_id_property = {
    "properties.serviceEndpoints": "service"
  }
}
```

### Resulting Resource JSON:

```json
{
  "body": {
    "properties": {
      "serviceEndpoints": [
        {
          "service": "Microsoft.Web"     // added by the provider
        },
        {
          "service": "Microsoft.KeyVaut" // ignored by provider because already present
        },
      ]
    }
  }
}
```

@jaredfholgate @ms-henglu

matt-FFFFFF:
@cwe1ss 

Thanks for adding your perspective. I think there are use cases for both:

1. Managing lists as sets (unordered), using a property as a unique key
2. Managing partial lists, also using a property as a unique key

The case for 2 is best illustrated with the diagnostic settings extension resource.

The API for each resource provides metrics and log categories. These are stored in a list, with the enabled property set to false. Unfortunately they are somewhat resource specific. That make them incredible difficult to manage using azapi, as the caller must specify the complete list.

In the case of these diagnostic settings categories, then I only want to cate about the setting that I want enabled.

Azure APIs are fun sometimes!

stemaMSFT:
Closing as completed — this was resolved in v2.9.0, which shipped `list_unique_id_property` and `ignore_other_items_in_list` fields for enhanced array/list change detection. Please reopen if there are any issues.

PR #1033, opened 2025-12-26
feat: add support for list_unique_id_property and ignore_other_items_in_list

kewalaka:
This change affects how JSON lists are compared and merged, it introduces two new properties `list_unique_id_property` and `ignore_other_items_in_list`.

The purpose of these two is to allow the following at specific paths:

* Allow list items to be identitied by a key (something other than `name` - to make the list matching stable under reordering or insertions)
* Ignore extra list items that appear in the remote (to preserve idempotency with state)

For example, taking service endpoints:

```hcl
list_unique_id_property = {
  "properties.serviceEndpoints" = "service"
}
```
... would set the key for `properties.serviceEndpoints` to `service`, matching values like:

```json
{
  "properties": {
    "serviceEndpoints": [
      { "service": "Microsoft.KeyVault", "locations": ["westus"] },
      { "service": "Microsoft.Web",     "locations": ["westus2"] }
    ]
  }
}
```

(another common example is `category` in diagnostic logs)

You can also ignore server-side additions for a specific path, e.g.:

```hcl
# ignore_other_items_in_list
ignore_other_items_in_list = [
  "properties.serviceEndpoints"
]
```

The recursive JSON traversal has been updated to allow this behaviour to be configured on specific paths.

### DIFF
--- a/docs/guides/feature_generate_config.md
+++ b/docs/guides/feature_generate_config.md
@@ -146,17 +146,17 @@ resource "azapi_resource" "mystorageaccount" {
   parent_id = "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/myResourceGroup"
   name      = "mystorageaccount"
   location  = "eastus"
-  
+
   body = {
     kind = "StorageV2"
     sku = {
       name = "Standard_LRS"
     }
     properties = {
-      accessTier                   = "Hot"
-      allowBlobPublicAccess        = false
-      minimumTlsVersion            = "TLS1_2"
-      supportsHttpsTrafficOnly     = true
+      accessTier               = "Hot"
+      allowBlobPublicAccess    = false
+      minimumTlsVersion        = "TLS1_2"
+      supportsHttpsTrafficOnly = true
       // ... other properties
     }
   }
--- a/internal/services/azapi_resource_test.go
+++ b/internal/services/azapi_resource_test.go
@@ -4360,7 +4360,7 @@ resource "azapi_resource" "vault" {
       sku = {
         family = "A"
         name   = "standard"
-      }      
+      }
       tenantId = data.azapi_client_config.current.tenant_id
     }
   }
@@ -4373,8 +4373,8 @@ resource "azapi_resource" "workspace" {
   location  = azapi_resource.resourceGroup.location
   body = {
     properties = {
-      sku                        = { name = "PerGB2018" }
-      retentionInDays            = 30
+      sku                             = { name = "PerGB2018" }
+      retentionInDays                 = 30
       publicNetworkAccessForIngestion = "Enabled"
       publicNetworkAccessForQuery     = "Enabled"
     }
@@ -4418,7 +4418,7 @@ resource "azapi_resource" "test" {
 }
 
 locals {
-  logs = azapi_resource.test.output.properties.logs
+  logs                = azapi_resource.test.output.properties.logs
   audit_event_enabled = try([for l in local.logs : l.enabled if l.category == "AuditEvent"][0], null)
 }
 
@@ -4466,8 +4466,8 @@ resource "azapi_resource" "test" {
 }
 
 locals {
-  logs = azapi_resource.test.output.properties.logs
-  audit_event_enabled = try([for l in local.logs : l.enabled if l.category == "AuditEvent"][0], null)
+  logs                                    = azapi_resource.test.output.properties.logs
+  audit_event_enabled                     = try([for l in local.logs : l.enabled if l.category == "AuditEvent"][0], null)
   azure_policy_evaluation_details_enabled = try([for l in local.logs : l.enabled if l.category == "AzurePolicyEvaluationDetails"][0], null)
 }
 
--- a/internal/services/azapi_update_resource_test.go
+++ b/internal/services/azapi_update_resource_test.go
@@ -1115,7 +1115,7 @@ resource "azapi_resource" "vault" {
       sku = {
         family = "A"
         name   = "standard"
-      }      
+      }
       tenantId = data.azapi_client_config.current.tenant_id
     }
   }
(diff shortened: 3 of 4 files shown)


kewalaka:
I forgot to tag the issue #995

utils/json.go
```diff
@@ -178,29 +216,43 @@ func UpdateObject(old interface{}, new interface{}, option UpdateJsonOption) int
...
 
-func identifierOfArrayItem(input interface{}) string {
+func identifierOfArrayItemByKey(input interface{}, key string) string {
 	inputMap, ok := input.(map[string]interface{})
 	if !ok {
 		return ""
 	}
-	name := inputMap["name"]
-	if name == nil {
+	if key == "" {
```
matt-FFFFFF: @kewalaka I think we should retain the previous functionality here. If key is not supplied, then try to use `name`, if it exists. If it does not then return empty string as before.

cwe1ss:
I still believe that `list_unique_id_property` should be something that's hard-coded in the provider. *Everyone* will have to set this to the same values for resource types with inconsistent lists (e.g. service endpoints). And it will be very hard for everyone to discover the resource types where it's necessary. You'll typically only discover this after a while when you get drift. If the list of cases were hard-coded in the provider, you would automatically make life easier for all users of the provider.

matt-FFFFFF:
Also: never let perfect be the enemy of the good. Merging this won't prevent implementation of the above suggestion. It's a step along the way.

kewalaka:
> It only helps fixing them faster for one user (which is nice and important, sure, but it doesn't help anyone else).

This is extremely useful for module authors - it is specifically helpful to simplify work in AVM modules, which currently need some hair-pulling HCL to pull off :-) - to the point that I'm not comfortable pushing the work and have been holding off using it commercially, whilst waiting for something like this to land.

Even for those not using AVM modules - the same abstraction applies - they can take the learnings from AVM and incorporate it into their own modules.

So it isn't value for one - it just allows saner abstractions at the module level, which unlocks work to use AzAPI more.

I see where you're coming from - and I grant that common problems should ideally be masked - but I can't find a good way to approach this that fits the azapi philosophy, open to suggestions.

kewalaka:
@matt-FFFFFF would it be acceptable to switch the property matching to use `jmes.search()` or would this open problems or be a performance concern?

Illustration - 

https://github.com/kewalaka/terraform-provider-azapi/pull/2/changes#diff-fcb2ff69e6cb2cd5cbcbc39806686890ebbfe8fcfa33c6ff9ff4eb6ed69bb1e1 


Noting i need to run `go generate` too for the unit test CI.

if the above looks ok i can fold it in here?

cwe1ss:
> I searched all list attributes in Azure resources with the latest stable/preview API versions, and found 3000+ records: https://gist.github.com/ms-henglu/2f719c3a7ce2ca697807ace052cf096e.

@ms-henglu this is very interesting, thank you for compiling the list. would it be possible to exclude any that have a `name`-field, because that case is already handled by the provider. That would then be the actual list in scope of this feature, right?

matt-FFFFFF:
@kewalaka I think the composite key idea is a good one and is a more robust implementation. Instead of using JMES, use [`strings.Split()`](https://pkg.go.dev/strings#Split) and test for each key, if length is 1 then retain as-is functionality. If null then empty string.

Then use [`strings.Join()`](https://pkg.go.dev/strings#Join) to create the composite key.

utils/json_test.go
```diff
@@ -1217,3 +1335,61 @@ func Test_FilterFields(t *testing.T) {
 		}
 	}
 }
+
+// Test_UpdateObject_CompositeKeyIdentifier tests composite keys for list item matching.
+func Test_UpdateObject_CompositeKeyIdentifier(t *testing.T) {
```
matt-FFFFFF: Make this table based with more comprehensive coverage

kewalaka:
was just doing a bit of interactive testing with a storage account diagnostics

https://github.com/kewalaka/azapi-demos/blob/main/tf-995-storagediags

All seems ok, one thing I did notice, it looks like there is a create bug where if there are no changes (e.g. if "logs" is empty), then it seems to get into a perpetual loop during creation.  I don't think that is related to this change as it is in the read-back - and if i remove the new flags the problem persists.

Apart from that things look good to me - would you like me to merge https://github.com/kewalaka/terraform-provider-azapi/pull/4 into here or are further discussions required first?

matt-FFFFFF:
@kewalaka do you have logs for the above issue? Can you throw them in a gist or something?

Assuming you are running with dev overrides?

ms-henglu:
Thanks @kewalaka for the private data solution, and thanks @matt-FFFFFF for the clarification.

But I still have questions about these cases:

> The current implementation will mean that a practitioner cannot delete items from an object array. As soon as they are removed from config, they will be seen as `other`, therefore won't be deleted.

It only happens when they enables the "ignore_other_items_in_list" feature:
For the AKS node pool, I'd recommend to use "ignore_other_items_in_list" in the managed cluster to ignore other pools, and use the child resources to manage the node pool.

For the AppGW and service endpoints, do they need to use the "ignore_other_items_in_list"? I thought they should manage all the items in Terraform, there should be no other items to ignore?

In summary, I think the issue only happens when users manage some of the items in Terraform, and the others are added by the API, so users enabled the "ignore_other_items_in_list" feature. And this kind of case is rare?

 If the unmanaged items are added by the users with another client tool, then this is a bad practice.

kewalaka:
Another scenario.  A module (e.g. avm interfaces) uses `ignore_other_items_in_list` to aid with idempotency.

A user has a resource (e.g. storage account) with specific diagnostics defined (e.g. read and write).  They decide read is unnecessary and remove it from the config.

This wont work, as it goes to "other". 

Instead they would have to either enable the provider feature, or specifically set the `enabled` flag to false.

It would be unintuitive, and potentially hidden because the decision to ignore list items would be a module concern.

matt-FFFFFF:
I get these may be niche things.

Let's merge and see what feedback we get.

ms-henglu:
Thanks @kewalaka for the explanation and examples!

ms-henglu:
Thanks @kewalaka and @matt-FFFFFF for contributing to this feature! I'll merge this PR.

Issue #1048, opened 2026-02-03
"204 No Content" response causes `azapi_data_plane_resource` to fail on update in-place


efrankenheimer:
Hello friends

When I run tf apply resulting in an update in-place action against an `azapi_data_plane_resource` utilizing the `Microsoft.Search/searchServices` API, a 204 is returned on success, which causes Terraform to mark the apply as failed.

Issue [440](https://github.com/Azure/terraform-provider-azapi/issues/440) discusses the same issue, but for `azapi_resource_action`.  Fix [441](https://github.com/Azure/terraform-provider-azapi/pull/441) added `http.StatusNoContent` to success response codes in [internal/clients/resource_client.go](https://github.com/Azure/terraform-provider-azapi/pull/441/changes#diff-9d255d61148b25074702079af37694d162a4968f9729e9fbe36651f76bbcf0d0).

The equivalent in [internal/clients/data_plane_client.go](https://github.com/Azure/terraform-provider-azapi/blob/cb25beafeb259fc5309f4fa4ee101519e803bea5/internal/clients/data_plane_client.go#L260) does not contain `http.StatusNoContent`, resulting in a `204` being treated as a failure.

Error encountered (initial creation works fine, modifications result in update in-place fail):

```
  │ Error: Failed to create/update resource
  │ 
  │   with azapi_data_plane_resource.default_datasource,
  │   on main.tf line 1241, in resource "azapi_data_plane_resource" "default_datasource":
  │ 1241: resource "azapi_data_plane_resource" "default_datasource" {
  │ 
  │ creating/updating "Resource: (ResourceId
  │ \"srch-d-eus2.search.windows.net/datasources('default-datasource')\"
  │ / Api Version \"2024-07-01\")": PUT
  │ https://srch-d-eus2.search.windows.net/datasources('default-datasource')
  │ --------------------------------------------------------------------------------
  │ RESPONSE 204: 204 No Content
  │ ERROR CODE UNAVAILABLE
  │ --------------------------------------------------------------------------------
  │ Response contained no body
  │ --------------------------------------------------------------------------------
```

As with Issue 440, it seems `azapi` should treat a `204` as succes.  [Azure AI Search HTTP status codes in the API reference](https://learn.microsoft.com/en-us/rest/api/searchservice/http-status-codes) show 204 as a "Success on PUT or POST".

[...]

PR #1055, opened 2026-03-03
bugfix: azapi_data_plane_resource doesn't support 204 status code

dinigo:
fix https://github.com/Azure/terraform-provider-azapi/issues/1048

Fix the issue when UPSERTING a dataplane resource when the resource already exists. In this case the API answers with a 204 instead of a 200/201.

```
╷
│ Error: Failed to create/update resource
│ 
│   with azapi_data_plane_resource.works_council_index,
│   on rag.tf line 20, in resource "azapi_data_plane_resource" "rag_index":
│   20: resource "azapi_data_plane_resource" "rag_index" {
│ 
│ creating/updating "Resource: (ResourceId \"rag-westeurope.search.windows.net/indexes('rag-index')\" / Api Version \"2025-09-01\")": PUT
│ https://rag-westeurope.search.windows.net/indexes('rag-index')
│ --------------------------------------------------------------------------------
│ RESPONSE 204: 204 No Content
│ ERROR CODE UNAVAILABLE
│ --------------------------------------------------------------------------------
│ Response contained no body
│ --------------------------------------------------------------------------------
│
```

This implements the same bugfix as https://github.com/Azure/terraform-provider-azapi/pull/441

### DIFF
--- a/internal/clients/data_plane_client.go
+++ b/internal/clients/data_plane_client.go
@@ -104,7 +104,7 @@ func (client *DataPlaneClient) CreateOrUpdateThenPoll(ctx context.Context, id pa
 	if err != nil {
 		return nil, err
 	}
-	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusCreated, http.StatusAccepted) {
+	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusCreated, http.StatusAccepted, http.StatusNoContent) {
 		return nil, runtime.NewResponseError(resp)
 	}
 


internal/clients/data_plane_client.go
```diff
@@ -257,7 +257,7 @@ func (client *DataPlaneClient) Action(ctx context.Context, resourceID string, ac
 	if err != nil {
 		return nil, err
 	}
-	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusCreated, http.StatusAccepted) {
+	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusCreated, http.StatusAccepted, http.StatusNoContent) {
```
ms-henglu: This is the Action method, I think you need to fix it in https://github.com/dinigo/terraform-provider-azapi/blob/0ace79be8f7bd5e0dbc366c3939aa0d94ae898d8/internal/clients/data_plane_client.go#L107

ms-henglu (changes requested):
Hi @dinigo , thank you so much for taking time to open this PR!

The fix looks good, but I think it should be added to the create/update method instead of the action method. Please check.

internal/clients/data_plane_client.go
dinigo: Thanks! I added it in the right block and it accepts the response correctly now

Issue #1043, opened 2026-01-30
Terraform import for Microsoft.Communication/communicationServices/smtpUsernames@2025-09-01 uses incorrect API version


trs1991:
When attempting to import a SMTP username resource the provider by default is trying to use an incorrect API version that doesn't exist for this resource which is api version 2025-05-01-preview:

`terraform import 'azapi_resource.smtp_username[0]' "/subscriptions/[REDACTED]/resourceGroups/[REDACTED]/providers/Microsoft.Communication/communicationServices/[REDACTED]/smtpUsernames/smtp-user"`

```
│ Error: Failed to retrieve resource
│
│ reading Resource: (ResourceId
│ "/subscriptions/[REDACTED]/resourceGroups/[REDACTED]/providers/Microsoft.Communication/communicationServices/[REDACTED]/smtpUsernames/smtp-user"
│ / Api Version "2025-05-01-preview"): GET
│https://management.azure.com/subscriptions/[REDACTED]/resourceGroups/[REDACTED]/providers/Microsoft.Communication/communicationServices/[REDACTED]/smtpUsernames/smtp-user"
│ --------------------------------------------------------------------------------
│ RESPONSE 400: 400 Bad Request
│ ERROR CODE: NoRegisteredProviderFound
│ --------------------------------------------------------------------------------
│ {
│   "error": {
│     "code": "NoRegisteredProviderFound",
│     "message": "No registered resource provider found for location 'global' and API version '2025-05-01-preview' for type 'CommunicationServices/SmtpUsernames'. The supported api-versions are '2024-09-01-preview, 2025-01-25-preview, 2025-09-01, 2025-09-01-preview'. The supported locations are 'global'."
│   }
│ }
│ --------------------------------------------------------------------------------
```

When passing in a valid API version, in this case 2025-09-01, the import is a success

```
terraform import 'azapi_resource.smtp_username[0]' "/subscriptions/[REDACTED]/resourceGroups/[REDACTED]/providers/Microsoft.Communication/communicationServices/[REDACTED]/smtpUsernames/smtp-user?api-version=2025-09-01"

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.
```

This is the terraform code used for the import:

```
resource "azapi_resource" "smtp_username" {
  count                     = 1
  type                      = "Microsoft.Communication/communicationServices/smtpUsernames@2025-09-01"
  location                  = "global"
[...]

PR #1093, opened 2026-04-09
`azapi_resource`: import: fallback to up to 3 most recent indexed api versions on 400 or 404 if it's not explicitly specified

EldarBorge:
Retry logic when API-versions in Azure have diverged from what's generated in the provider.
Will try the older api version until it gets success.

The motivation is to close and prevent issues like #1089 and #1043. 

Not tested.

**AI disclaimer**: Initial code made by opus 4.6 with github copilot extension in vscode, then manually edited by me.


gerrytan:
Thank you @EldarBorge , but the API version to be used for importing resources should be explicitly defined in configuration right?

If a preview version no longer available, then you should update your Terraform config to import using the correct one.

See: https://registry.terraform.io/providers/Azure/azapi/latest/docs/resources/resource#import

Note how the vnet id in the example below can accept `api-version` query param:

```
import {
  to = azapi_resource.example
  id = "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/example-rg/providers/Microsoft.Network/virtualNetworks/example-vnet?api-version=2023-11-01"
}
```

gerrytan:
Thank you for the contribution @EldarBorge , after discussing with @magodo, we have decided the "make api-version required in import id" breaking change is too painful for users, so we're going to adopt your idea.

I have pushed some refinement commits and ran acctests: https://dev.azure.com/azclitools/internal/_build/results?buildId=326636&view=results (access restricted).

The retry logic is dependant on 400 and 404 errors, and capped at maximum 3 most recent indexed versions. I've also added markdown documentation to explain the behaviour.

@magodo can you please review this PR.

.github/workflows/unit-test.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
       - run: go generate ./...
       - name: git diff
         run: |
-          git diff --exit-code || \
+          git diff --exit-code --ignore-blank-lines --ignore-space-at-eol || \
```
magodo: Why are these options needed given they are run in CI?

docs/resources/resource.md
```diff
@@ -338,3 +338,7 @@ Required:
 Optional:
 
 - `type` (String) The Azure resource type.
+
+### API Version for Initial Import Call
+
+While processing an import request, the provider cannot see the target HCL configuration, so it cannot determine which api-version to use for the initial import GET call. To work around this, the provider falls back to the 3 most recent [indexed](https://github.com/Azure/terraform-provider-azapi/blob/main/internal/azure/generated/index.json) api-versions (regardless of preview / stable), trying each in turn. The provider treats any 400 or 404 response as a signal to attempt the next older version. This fallback can result in an api-version mismatch; to avoid it, specify an explicit api-version using the `api-version` query parameter or the resource `type` attribute.
```
magodo: You shall omit the HCL config part as import doesn't need HCL config at all. Also, you shall mention that this behavior only applies when `api-version` is not specified.

magodo (commented):
Some minor comments, otherwise LGTM!

.github/workflows/unit-test.yaml
gerrytan: Previously this was done to avoid perpetual diff because locally I have a terrafmt precommit git hook re-ading newline at eol. But you are right I could check in the commit without the precommit hook instead. This diff has been reverted.

docs/resources/resource.md
gerrytan: Rephrased into

> While processing an import request, the provider cannot determine which api-version to use...

magodo: Shouldn't this be changed in the resource's source file (the template)?

gerrytan:
Thanks for your feedbacks @magodo , they have been addressed. acctest result is good. Please take another look at this PR:

image

magodo (approved):

PR #1100, opened 2026-04-20
Fix dependabot.yml syntax error

gerrytan:
The dependabot.yml contains syntax error which causes automatic update to not trigger.

https://github.com/Azure/terraform-provider-azapi/network/updates

image

Refer to https://www.schemastore.org/dependabot-2.0.json: the doc on `vendor` says:

> Don't use this option if you're using 'gomod'.

## Testing

Have checked local validator pass. Will check the dependency setting section above again after merging to main.


wuxu92 (approved):
LGTM!

PR #1122, opened 2026-05-15
Add `preserve_resource_id_casing` provider feature flag

### LINKED ISSUE #1120: Provider feature flag to preserve existing resource ID casing in state
### Background

The Azure Verified Module `terraform-azurerm-avm-res-web-serverfarm` v2.0.0 switched from `azurerm_service_plan` to `azapi_resource`. Users upgrading from v1.0.0 hit `Unexpected Identity Change` errors because:

- The migrated state contained a resource ID with `serverFarms` (capital `F`), as written by the `azurerm` provider.
- `azapi_resource` (with `type = "Microsoft.Web/serverfarms@2025-03-01"`) recomputes the ID with `serverfarms` (lowercase `f`).
- Both spellings reference the same Azure resource, but the Terraform identity comparison is case-sensitive and fails.

Upstream report: Azure/terraform-azurerm-avm-res-web-serverfarm#122

### Proposal

Add an opt-in provider feature flag (default off) that, when enabled, preserves the existing resource ID casing in state if the value the provider would write back differs from state only by casing.

- Provider attribute: `preserve_resource_id_casing` (Boolean, `Optional`, default `false`).
- Environment variable: `ARM_PRESERVE_RESOURCE_ID_CASING`.
- Scope: only the `id` and `resource_id` attributes (and the framework identity that is derived from them). Other properties such as `type` are unaffected, so the user's HCL `type` value continues to drive Terraform state directly.

### Behaviour

When the flag is `true`:

- During Read and CreateUpdate, if `strings.EqualFold(state.id, computed.id)` and `state.id != computed.id`, the existing `state.id` casing is kept.
- Same for `resource_id`.

When the flag is `false` (default), behaviour is unchanged.

### Why opt-in

Casing preservation hides a real (if benign) difference between state and the ID the provider would otherwise compute. Defaulting it off keeps current behaviour for everyone and lets module authors / consumers turn it on when they need ID stability across migrations.

jaredfholgate:
## Summary

Adds an opt-in provider feature flag `preserve_resource_id_casing` (default `false`) that preserves the existing casing of `id` and `resource_id` in state when the value the provider would otherwise write back differs only by casing.

Fixes #1120 (which references the original report at Azure/terraform-azurerm-avm-res-web-serverfarm#122).

## Why

One example of the issue: The Azure Verified Module `terraform-azurerm-avm-res-web-serverfarm` v2.0.0 switched from `azurerm_service_plan` to `azapi_resource`. Users upgrading from v1.0.0 hit `Unexpected Identity Change` errors because `azurerm` had written `serverFarms` (capital `F`) into state, while `azapi_resource` recomputes the ID with the casing from the user's `type` attribute, which the AVM module sets to `Microsoft.Web/serverfarms@2025-03-01` (lowercase `f`). Both spellings reference the same Azure resource, but Terraform's identity comparison is case-sensitive.

## Behaviour

- **Default (`false`)**: identical to today; the provider writes `id` / `resource_id` as computed.
- **`true`**: in Read and CreateUpdate, if `strings.EqualFold(state.id, computed.id)` and they differ, the existing casing in state is kept. Same for `resource_id`.
- Genuinely different values (not just casing) always pass through unchanged.
- Scope is limited to the `id` and `resource_id` attributes (and the framework identity derived from them). The user's HCL `type` value is never silently overridden.

## Configuration

```hcl
provider "azapi" {
  preserve_resource_id_casing = true
}
```

or via env: `ARM_PRESERVE_RESOURCE_ID_CASING=true`.

## Tests

- Added unit test `TestPreserveCasing` in `internal/services/common_test.go` covering: disabled, empty existing, case-only difference, identical, genuinely different.
- I did not add an acceptance test: the casing scenario is hard to provoke from a fresh `terraform apply` (state ID matches user's `type` casing on first write); reproducing it requires a state migrated/imported from another source. Happy to add one (e.g. via `terraform import` with a re-cased URL) if reviewers prefer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ ENHANCEMENTS:
 - Bump Go version to 1.25.8 (GH-1082).
 - Bump `terraform-plugin-framework` from v1.16.1 to v1.19.0 (GH-1081).
 - Add example for `Microsoft.RedHatOpenShift/openShiftClusters` (GH-1083).
+- `azapi` provider: Support `preserve_resource_id_casing` feature flag (default `false`). When enabled, if the resource ID the provider would write back to state differs from the existing state value only by casing, the existing casing is preserved. This avoids spurious `Unexpected Identity Change` errors and diffs when consumers (or upstream modules) rely on a specific casing the Azure API may not preserve. Only the `id` and `resource_id` attributes are affected. Can also be sourced from the `ARM_PRESERVE_RESOURCE_ID_CASING` environment variable (GH-1122).
 
 BUG FIXES:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ ENHANCEMENTS:
 - Acquire a policy token when a request is blocked by an invoke policy (GH-1145).
 - Update bicep types to Azure/azure-rest-api-specs revision 51d53915b5f31b10e6645c136807b9d95f9f09d1 (GH-1159).
 - Improve the bicep types update process (GH-1155).
+- `azapi` provider: Support `preserve_resource_id_casing` feature flag (default `false`). When enabled, if the resource ID the provider would write back to state differs from the existing state value only by casing, the existing casing is preserved. This avoids spurious `Unexpected Identity Change` errors and diffs when consumers (or upstream modules) rely on a specific casing the Azure API may not preserve. Only the `id` and `resource_id` attributes are affected. Can also be sourced from the `ARM_PRESERVE_RESOURCE_ID_CASING` environment variable (GH-1122).
 
 BUG FIXES:
 
@@ -26,7 +27,6 @@ ENHANCEMENTS:
 - Bump Go version to 1.25.8 (GH-1082).
 - Bump `terraform-plugin-framework` from v1.16.1 to v1.19.0 (GH-1081).
 - Add example for `Microsoft.RedHatOpenShift/openShiftClusters` (GH-1083).
-- `azapi` provider: Support `preserve_resource_id_casing` feature flag (default `false`). When enabled, if the resource ID the provider would write back to state differs from the existing state value only by casing, the existing casing is preserved. This avoids spurious `Unexpected Identity Change` errors and diffs when consumers (or upstream modules) rely on a specific casing the Azure API may not preserve. Only the `id` and `resource_id` attributes are affected. Can also be sourced from the `ARM_PRESERVE_RESOURCE_ID_CASING` environment variable (GH-1122).
 
 BUG FIXES:
 
--- a/internal/services/common_test.go
+++ b/internal/services/common_test.go
@@ -1,0 +1,93 @@
+package services
+
+import "testing"
+
+func TestPreserveCasing(t *testing.T) {
+	const (
+		serverFarmsID = "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg1/providers/Microsoft.Web/serverFarms/plan1"
+		serverfarmsID = "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg1/providers/Microsoft.Web/serverfarms/plan1"
+	)
+
+	testCases := []struct {
+		name      string
+		existing  string
+		candidate string
+		enabled   bool
+		expected  string
+	}{
+		{
+			name:      "disabled returns candidate",
+			existing:  "/foo/Bar",
+			candidate: "/foo/bar",
+			enabled:   false,
+			expected:  "/foo/bar",
+		},
+		{
+			name:      "enabled with empty existing returns candidate",
+			existing:  "",
+			candidate: "/foo/bar",
+			enabled:   true,
+			expected:  "/foo/bar",
+		},
+		{
+			name:      "enabled with case-insensitive match keeps existing",
+			existing:  "/foo/Bar",
+			candidate: "/foo/bar",
+			enabled:   true,
+			expected:  "/foo/Bar",
+		},
+		{
+			name:      "enabled with identical values keeps existing",
+			existing:  "/foo/bar",
+			candidate: "/foo/bar",
+			enabled:   true,
+			expected:  "/foo/bar",
+		},
+		{
+			name:      "enabled with genuinely different values returns candidate",
+			existing:  "/foo/bar",
+			candidate: "/foo/baz",
+			enabled:   true,
+			expected:  "/foo/baz",
+		},
+		{
+			// Regression test for #1120: after a state migration the resource id
+			// may have been normalised to lower case (serverfarms) while the value
+			// the provider recomputes preserves the original casing (serverFarms).
+			// With the feature enabled the state casing must be preserved so the
+			// azapi_resource identity does not report a spurious change.
+			name:      "enabled preserves migrated serverFarms casing (#1120)",
+			existing:  serverFarmsID,
+			candidate: serverfarmsID,
+			enabled:   true,
+			expected:  serverFarmsID,
+		},
+		{
+			// Same scenario but with the feature disabled (the default): the
+			// recomputed candidate wins so behaviour is unchanged for existing users.
+			name:      "disabled does not preserve migrated serverFarms casing (default)",
+			existing:  serverFarmsID,
+			candidate: serverfarmsID,
+			enabled:   false,
+			expected:  serverfarmsID,
+		},
+		{
+			// azapi_update_resource.resource_id path: the bare Azure resource id
+			// (without the api-version suffix) must also keep its state casing.
+			name:      "enabled preserves resource_id path casing",
+			existing:  "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/RG1/providers/Microsoft.Web/serverFarms/plan1",
+			candidate: "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/rg1/providers/Microsoft.Web/serverfarms/plan1",
+			enabled:   true,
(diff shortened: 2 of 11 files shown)


jaredfholgate:
@ms-henglu Any thoughts on this and whether we could get this feature into the provider soon? Many thanks.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ ENHANCEMENTS:
 - Bump Go version to 1.25.8 (GH-1082).
 - Bump `terraform-plugin-framework` from v1.16.1 to v1.19.0 (GH-1081).
 - Add example for `Microsoft.RedHatOpenShift/openShiftClusters` (GH-1083).
+- `azapi` provider: Support `preserve_resource_id_casing` feature flag (default `false`). When enabled, if the resource ID the provider would write back to state differs from the existing state value only by casing, the existing casing is preserved. This avoids spurious `Unexpected Identity Change` errors and diffs when consumers (or upstream modules) rely on a specific casing the Azure API may not preserve. Only the `id` and `resource_id` attributes are affected. Can also be sourced from the `ARM_PRESERVE_RESOURCE_ID_CASING` environment variable.
```
hqhqhqhqhqhqhqhqhqhqhq: Small changelog issue from the stale branch: current main already has a GH-1125 bug-fix entry under v2.10.0. Could you rebase this and keep that entry, then add this feature under ENHANCEMENTS?

internal/services/common_test.go
```diff
@@ -0,0 +1,58 @@
+package services
+
+import "testing"
+
+func TestPreserveCasing(t *testing.T) {
```
hqhqhqhqhqhqhqhqhqhqhq, line 5: This helper test is a good start, but I think we should also cover the actual issue path from #1120.

The important case is migrated state with `serverFarms`, while `azapi_resource` recomputes `serverfarms`. With `preserve_resource_id_casing = true`, we should verify the old casing is kept in state/identity. With the default `false`, behavior should stay unchanged.

It would also be good to cover `azapi_update_resource.resource_id`, since this PR changes that path too.

hqhqhqhqhqhqhqhqhqhqhq (commented):
Thanks for working on this. I think the direction makes sense, but the branch looks stale against current main now.

Could you rebase it and make sure the new `preserve_resource_id_casing` flag is added alongside the existing `always_acquire_policy_token` provider flag? Right now the PR branch is missing that newer provider option from main, so we need to make sure the rebase keeps both.

CHANGELOG.md
jaredfholgate: Done — rebased onto current `main`, kept the existing `GH-1125` bug-fix entry under v2.10.0, and added `preserve_resource_id_casing` under `ENHANCEMENTS` as `(GH-1122)`.

internal/services/common_test.go
jaredfholgate: Good call — expanded `TestPreserveCasing` to cover the actual #1120 path: a migrated `serverFarms` value vs a recomputed `serverfarms` candidate now asserts the old casing is preserved when `preserve_resource_id_casing = true`, and that the default (`false`) leaves behavior unchanged. I also added a case for the `azapi_update_resource.resource_id` path since this PR touches it too. All 8 cases pass.

jaredfholgate:
Thanks for the review @hqhqhqhqhqhqhqhqhqhqhq — I've pushed an update addressing all three points:

- **Rebased onto current `main`**, so `preserve_resource_id_casing` now sits alongside the existing `always_acquire_policy_token` provider flag. Both flags are wired independently through the schema, `Configure`, and their env vars (`ARM_PRESERVE_RESOURCE_ID_CASING` / `ARM_ALWAYS_ACQUIRE_POLICY_TOKEN`), so the rebase keeps both. The branch is no longer conflicting.
- **CHANGELOG**: kept the existing `GH-1125` bug-fix entry under v2.10.0 and added this feature under `ENHANCEMENTS` as `(GH-1122)`.
- **Tests**: expanded `TestPreserveCasing` to cover the actual #1120 scenario (migrated `serverFarms` vs recomputed `serverfarms` — old casing kept when enabled, unchanged when default) and the `azapi_update_resource.resource_id` path.

`go build ./...` and `go test ./internal/services/...` pass locally.

internal/services/common_test.go
jaredfholgate: Follow-up: I've now also added resource-level acceptance tests in `internal/services/azapi_update_resource_preserve_resource_id_casing_test.go`:

- `TestAccGenericUpdateResource_preserveResourceIDCasing` — a fresh apply that drives the flag end-to-end through `azapi_update_resource.resource_id` (source resource declared as lower-case `webPubSub`, update resource as canonical `WebPubSub`) and asserts no perpetual diff.
- `TestAccAzapiUpdateResourceUpgrade_preserveResourceIDCasing` — a faithful reproduction of #1120: deploy with a previous provider version, then upgrade to this build with `preserve_resource_id_casing = true` and assert the plan stays empty (the previously stored casing is preserved rather than re-normalised).

Note: on a fresh create the prior state is empty, so the flag is only observable across a provider-version state migration — hence the upgrade test is the one that isolates the new behaviour. These are gated behind `TF_ACC` + live Azure like the other acceptance tests.

cwe1ss:
I'm wondering if this should really be a provider-level flag? We're managing thousands of resources using AzAPI, and suddenly, due to one specific resource type having a different behavior, we would need to switch a provider-flag, which then potentially affects all resources? It also probably means, that everyone who needs this flag will first run into this issue (because the flag is not set by default), they will then have to research the issue, hopefully find this flag, and activate it. This isn't a good user experience in my opinion.

I strongly believe that even though AzAPI is a generic provider, it should still be able to handle some known special cases automatically, without the user having to run into the issues first.

jaredfholgate:
> I'm wondering if this should really be a provider-level flag? We're managing thousands of resources using AzAPI, and suddenly, due to one specific resource type having a different behavior, we would need to switch a provider-flag, which then potentially affects all resources? It also probably means, that everyone who needs this flag will first run into this issue (because the flag is not set by default), they will then have to research the issue, hopefully find this flag, and activate it. This isn't a good user experience in my opinion.
> 
> I strongly believe that even though AzAPI is a generic provider, it should still be able to handle some known special cases automatically, without the user having to run into the issues first.

@hqhqhqhqhqhqhqhqhqhqhq I kind of agree with this. I was being cautious by not changing the default behaviour, but realistically having this on by default would only improve the experience for most users. It doesn't have any impact on resources that are not affected by this problem, so safe to default to `true` if you want to go ahead with that change?

hqhqhqhqhqhqhqhqhqhqhq:
Thanks for following up on this. Since the branch was last updated, #1156 has merged and changed how `azapi_resource` sets its identity during `Read`.

Before we decide whether this flag should default to `true`, could you rebase once more and check whether the original #1120 migration still fails on current `main`?

Ideally, the test should cover the actual migration from `azurerm_service_plan` to `azapi_resource`, where the existing state contains `serverFarms` but the AzAPI configuration uses `serverfarms`. The current acceptance test uses `azapi_update_resource` and WebPubSub, so it doesn’t quite reproduce that original path.

If current `main` already fixes it, we may not need the new provider setting. If it still fails, having that exact regression will help us decide whether casing preservation should be automatic or configurable.

jaredfholgate:
> Thanks for following up on this. Since the branch was last updated, #1156 has merged and changed how `azapi_resource` sets its identity during `Read`.
> 
> Before we decide whether this flag should default to `true`, could you rebase once more and check whether the original #1120 migration still fails on current `main`?
> 
> Ideally, the test should cover the actual migration from `azurerm_service_plan` to `azapi_resource`, where the existing state contains `serverFarms` but the AzAPI configuration uses `serverfarms`. The current acceptance test uses `azapi_update_resource` and WebPubSub, so it doesn’t quite reproduce that original path.
> 
> If current `main` already fixes it, we may not need the new provider setting. If it still fails, having that exact regression will help us decide whether casing preservation should be automatic or configurable.

@hqhqhqhqhqhqhqhqhqhqhq (your GitHub handle is great, I constantly think there is some error when I see it... :))

Rebased onto main. #1156 doesn't fix this issue so far as I can see in my analysis and testing. I've added `TestPreserveResourceIDCasingAzapiResource`, a plain unit test over the real `parse.NewResourceID`, so the flip is verifiable without `TF_ACC`.

On the flag: my personal view is it should be on by default to avoid friction for customers, but will let you decide. Keep the flag to turn it off for any edge cases we haven't considered.

hqhqhqhqhqhqhqhqhqhqhq (approved):

hqhqhqhqhqhqhqhqhqhqhq:
> > Thanks for following up on this. Since the branch was last updated, #1156 has merged and changed how `azapi_resource` sets its identity during `Read`.
> 
> > 
> 
> > Before we decide whether this flag should default to `true`, could you rebase once more and check whether the original #1120 migration still fails on current `main`?
> 
> > 
> 
> > Ideally, the test should cover the actual migration from `azurerm_service_plan` to `azapi_resource`, where the existing state contains `serverFarms` but the AzAPI configuration uses `serverfarms`. The current acceptance test uses `azapi_update_resource` and WebPubSub, so it doesn’t quite reproduce that original path.
> 
> > 
> 
> > If current `main` already fixes it, we may not need the new provider setting. If it still fails, having that exact regression will help us decide whether casing preservation should be automatic or configurable.
> 
> 
> 
> @hqhqhqhqhqhqhqhqhqhqhq (your GitHub handle is great, I constantly think there is some error when I see it... :))
> 
> 
> 
> Rebased onto main. #1156 doesn't fix this issue so far as I can see in my analysis and testing. I've added `TestPreserveResourceIDCasingAzapiResource`, a plain unit test over the real `parse.NewResourceID`, so the flip is verifiable without `TF_ACC`.
> 
> 
> 
> On the flag: my personal view is it should be on by default to avoid friction for customers, but will let you decide. Keep the flag to turn it off for any edge cases we haven't considered.

Thanks, looks good to me and approved!

PR #1079, opened 2026-03-22
feat(client): return last retryable error instead of context deadline exceeded

matt-FFFFFF:
## Summary

- When the retry loop exhausts the context deadline (timeout), capture and surface the **last retryable error** (with full ARM error text: status code, error code, JSON message body) instead of the unhelpful `context deadline exceeded` message.
- Fix pre-existing bug where `ShouldRetry` regex-matched successful (2xx) responses, causing infinite retries with `error_message_regex = [".*"]`.

### Before (v2.8)

The user only sees `context deadline exceeded` with no indication of *why* the retries were happening:

```
╷
│ Error: Failed to retrieve resource
│
│   with data.azapi_resource.rg_not_exist,
│   on main.tf line 48, in data "azapi_resource" "rg_not_exist":
│   48: data "azapi_resource" "rg_not_exist" {
│
│ retrieving resource "Resource: (ResourceId
│ "/subscriptions/57934baa-70ec-4d1a-95e0-3799a838811f/resourceGroups/rg-charming-stingray"
│ / Api Version "2025-04-01")": context deadline exceeded
╵
```

### After (this PR)

The full ARM error is surfaced — status code, error code, and JSON message body:

```
╷
│ Error: Failed to retrieve resource
│
│   with data.azapi_resource.rg_not_exist,
│   on main.tf line 48, in data "azapi_resource" "rg_not_exist":
│   48: data "azapi_resource" "rg_not_exist" {
│
│ retrieving resource "Resource: (ResourceId
│ "/subscriptions/57934baa-70ec-4d1a-95e0-3799a838811f/resourceGroups/rg-internal-dog"
│ / Api Version "2025-04-01")": context deadline exceeded, last retryable
│ error: GET
│ https://management.azure.com/subscriptions/57934baa-70ec-4d1a-95e0-3799a838811f/resourceGroups/rg-internal-dog
│ --------------------------------------------------------------------------------
│ RESPONSE 404: 404 Not Found
│ ERROR CODE: ResourceGroupNotFound
│ --------------------------------------------------------------------------------
│ {
│   "error": {
│     "code": "ResourceGroupNotFound",
│     "message": "Resource group 'rg-internal-dog' could not be found."
│   }
│ }
│ --------------------------------------------------------------------------------
│
╵
```

## Implementation

- Add `LastRetryError` struct (thread-safe) with `Set`/`Get` and `WrapContextError` helper in `internal/clients/last_retry_error.go`
[...]

PR #1141, opened 2026-06-13
feat(data-plane): add Azure Table Storage support

kewalaka:
## Summary

Adds Azure Table Storage support to the existing AzAPI data-plane resource, data source, and ephemeral resource, using the existing customization framework.

This PR no longer introduces the generic `data.azapi_data_plane_resource` data source: that work, together with `ephemeral.azapi_data_plane_resource`, was merged separately in #1199. The optional-name/singleton fix for #1091 was merged in #1178.

## Current scope

| Resource type suffix | Behavior |
| --- | --- |
| `Microsoft.Storage/storageAccounts/tableServices/tables` | Create, read, and delete tables. Table updates are not supported. |
| `Microsoft.Storage/storageAccounts/tableServices/tables/entities` | Create/update entities using PUT (insert-or-replace), read an entity, and delete it. |
| `Microsoft.Storage/storageAccounts/tableServices/tables/entitiesCollection` | Read an entity collection through the data source, with optional query parameters. |

- Entity identity is supplied directly in `parent_id`, ending in `tableName(PartitionKey='...',RowKey='...')`; no separate `name` or `identifiers` attribute is introduced for entities.
- Storage authentication audiences/endpoints are registered for Azure public, US Government, and China clouds.
- The data source and ephemeral resource gain generic `headers` and `query_parameters` attributes, using existing request-option helpers and the same schema pattern as `data.azapi_resource`. Query parameters allow service-side collection filtering such as `$filter`; there is no Table-specific query schema.
- The existing ID parser gains support for embedded `{parentId}` placeholders, used by the collection URL format `{parentId}()`.
- Managed-resource planning rejects the read-only collection type and directs users to the data source.
- `DataPlaneClient.Action` accepts HTTP 204, as returned by Table entity insert-or-replace operations.
- Adds Table Storage examples, generated documentation, and parser, validation, customization, and acceptance-test coverage.
- Entity reads preserve the complete API response, including identity fields, timestamps, and OData metadata.
- Entity writes replace the stored entity's user-defined properties. Properties omitted from a write are removed, including externally added properties. This does not introduce additional drift detection for unconfigured properties.
- Existing `sensitive_body_version` filtering is unchanged: unchanged sensitive properties omitted from a replacement request can be removed. This limitation is documented in the example and generated resource documentation.


LuuOW (commented):
Technical audit: code implementation verified for system consistency.

kewalaka:
thanks for the feedback @LuuOW - i've found some resource-specific issues hence converting this to draft, good to know i'm heading in the right direction from a consistency point of view.

i hope to get an update in the next day or so.

kewalaka:
happy with this now

example of this reading and writing entities from table storage:

https://github.com/kewalaka/terraform-azapi-global-outputs

kewalaka:
a demo to illustrate it in use:

https://github.com/kewalaka/terraform-azapi-global-outputs

This uses table storage as a configuration store, the CI builds the AzAPI provider from my branch and exercises the example in this repo.

kewalaka:
oops missed the fmt and linting checks - sorry folks - just committed.

kewalaka:
re-aligned with changes in #1079, tests & lint passes.  Spotted China endpoints were missed, also ran gen-doc.

besides the standard acc tests, re-ran it on my demo too:

https://github.com/kewalaka/terraform-azapi-global-outputs/actions/runs/29121317959

internal/clients/data_plane_client.go
```diff
@@ -75,13 +75,25 @@ func (client *DataPlaneClient) CreateOrUpdateThenPoll(ctx context.Context, id pa
 	if options.RetryOptions != nil {
 		ctx = policy.WithRetryOptions(ctx, *options.RetryOptions)
 	}
+
+	// Inject headers into context so they are preserved across poller requests
+	ctxHeaders := http.Header{}
```
LingyuTang: This change is great improvement. Only thing is that few lines below it still trying to set the header, might be good to remove those for removing duplicates.

internal/services/parse/data_plane_resource_id.go
```diff
@@ -125,3 +97,96 @@ func (id DataPlaneResourceId) String() string {
 func (id DataPlaneResourceId) ID() string {
 	return id.AzureResourceId
 }
+
+func renderDataPlaneURLFormat(urlFormat string, values map[string]string) (string, error) {
```
LingyuTang, line 104: The change to use regex is nice. Could you add some unit test for this to make sure it works as expected?

internal/services/parse/data_plane_resource_id.go
```diff
@@ -125,3 +97,96 @@ func (id DataPlaneResourceId) String() string {
...
+		}
+		return value
+	})
+	if len(missing) != 0 {
+		return "", fmt.Errorf("missing required data plane identifiers for URL format %q: %s", urlFormat, strings.Join(missing, ", "))
+	}
+	return rendered, nil
+}
+
+func parseDataPlaneURLFormat(urlFormat string, actual string) (map[string]string, error) {
```
LingyuTang, line 134: Nice work! Just need some unit test for this helper functions.

internal/services/parse/data_plane_type_map.go
```diff
@@ -33,3 +38,56 @@ func findApiPathByResourceType(resourceType string) *ApiPath {
...
+
+var dataPlanePlaceholderPattern = regexp.MustCompile(`\{([^{}]+)\}`)
+
+type dataPlanePlaceholder struct {
+	Key          string
+	DefaultValue string
+	HasDefault   bool
+}
+
+func placeholdersForURLFormat(urlFormat string) []dataPlanePlaceholder {
```
LingyuTang, line 50: Probobaly add some unit tests for this helper function as well.

internal/services/parse/data_plane_resource_id.go
```diff
@@ -13,37 +14,36 @@ type DataPlaneResourceId struct {
 	AzureResourceType string
 	Name              string
 	ParentId          string
+	Identifiers       map[string]string
```
LingyuTang: Seems like most of the data plane source only using Name and Identifiers are just for Storage, as it requires partition key and row key together. So might be good to add a comment here to explian why we need it.

internal/services/azapi_data_plane_resource_name_validation_test.go
```diff
@@ -4,17 +4,18 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
-func TestValidateDataPlaneResourceName(t *testing.T) {
+func TestValidateDataPlaneResourceAddress(t *testing.T) {
```
LingyuTang, line 11: Instead of using `address` might be better renaming to identifiers? Also the test file name should be changed as well.

internal/services/customization/storage_table_entity_customization.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+		if strings.HasPrefix(key, "odata.") {
+			continue
+		}
+		flattened[key] = value
+	}
+	return flattened, nil
+}
+
+const httpMethodMerge = "MERGE"
```
LingyuTang, line 155: The merge behavior is required in reasonable context but will be good to add a note in docuementation as it is different from others

internal/services/azapi_data_plane_resource_common.go
```diff
@@ -0,0 +1,113 @@
...
+		return nil
+	}
+	if (*customizedResource).CreateFunc() == nil {
+		typeBase := strings.Split(resourceType, "@")[0]
+		return fmt.Errorf(`resource type %q does not support create/update/delete; use data.azapi_data_plane_resource to read it instead`, typeBase)
+	}
+	return nil
+}
+
+func validateDataPlaneResourceAddress(config *DataPlaneResourceModel) error {
```
LingyuTang, line 30: validateDataPlaneResourceAddress sounds a little bit vague, might be good to change to `validateDataPlaneResourceNameAndIdentifier` or others.

LingyuTang (changes requested):
Thanks for the PR! Overall, this is great work and in a right direction. I've left a few comments here and there with some minor suggestions for improvement. One thing we feel is important to include as part of this change is support for the ephemeral feature, shipped alongside the data source. Given the security implications of persisting read-back data in the Terraform state, it would be good to address this from day one rather than adding it later.

If you're happy with it, I can also take over and help with the further work.

One other thing: it would be good to run go generate again, as one of the pipelines is currently failing and may just need the generated docs to be refreshed.

kewalaka:
Thanks.  Travelling for a while but happy to support where I can

LingyuTang:
Hey, we've merged PRs #1199 and #1178 to rescope this PR to focus only on Table Storage/identifier support and the request header improvements as the Table Storage/identifier support changes are more significant and require additional discussion to determine the best long-term approach.

I'll handle the merge conflicts, resolve the comments and update this branch shortly. Thanks.

kewalaka:
Thanks.  Apologies for going quiet I've been travelling without access to a screen.

Appreciate the help to get this landed.

LingyuTang:
Hey @kewalaka, thanks again for contributing to the codebase.

We had an internal discussion about Table Storage support. While adding an identifiers map is a reasonable approach, AzAPI aims to stay as thin a layer over the Azure API as possible, so we'd prefer to avoid introducing resource-specific properties when the same scenario can be handled generically.

For this case, we think the composite key could be included directly in `parent_id`, for example:

`parent_id = "${azapi_resource.storageAccount.name}.table.core.windows.net/${azurerm_storage_table.example.name}(PartitionKey='example',RowKey='state')"`

Regarding `DisableAPIVersionQueryParameter` and `APIVersionHeaderName`, we'd prefer to reuse the existing request header mechanisms (read/create/update/delete_headers) rather than add new options. Since Azure Storage APIs currently accept `api-version` as a query parameter, we don't see a strong need for `DisableAPIVersionQueryParameter` at this time.

I realize this is a fairly significant change in direction from the current implementation. If you're happy to continue with this PR, we'd suggest:

- representing the entity key through parent_id rather than introducing identifiers;
- reusing the existing request header mechanisms instead of adding APIVersionHeaderName;
- omitting DisableAPIVersionQueryParameter for now.

If you'd prefer, I'm also happy to take over and rework these pieces. Please let us know if you have any concerns or questions.

kewalaka:
All makes sense will take a look at this properly when I return in a week or so.

kewalaka:
> I realize this is a fairly significant change in direction from the current implementation. If you're happy to continue with this PR, we'd suggest:
> 
>     * representing the entity key through parent_id rather than introducing identifiers;
> 
>     * reusing the existing request header mechanisms instead of adding APIVersionHeaderName;
> 
>     * omitting DisableAPIVersionQueryParameter for now.
> 

- Removed the proposed identifiers attribute and generalized identifier-template machinery; composite keys now live in parent_id.
 
 - Removed APIVersionHeaderName and the automatic Storage-specific header injection. Examples supply x-ms-version and OData headers through existing managed-resource create_headers, read_headers, update_headers, and delete_headers, and the data source's generic headers.
 
- Removed DisableAPIVersionQueryParameter; the client's existing api-version query parameter remains unchanged.

The latest commit, 06ade641896601e89464eb0c2ba6b0336fa02eca, completes the header pass-through change by removing the Storage header helpers and implicit If-Match default. Entity deletion examples explicitly supply If-Match: * through delete_headers.

kewalaka:
I’ve changed Table entity writes from MERGE to PUT.

With MERGE, removing a property from Terraform configuration leaves that property in the stored entity. PUT makes property removal effective without adding property tracking or reconciliation logic to AzAPI.

`entitiesCollection` adds some complexity and I'm unsure if it is worth it, I am mirroring the surface of AzureRM (`azurerm_storage_table_entities`).

I note that #1199 removed support for the request options deferring it until it is required, headers are needed for table entity data sources, query parameters only for the collections support.

I'm not entirely happy with the placement of the second test in the ephemeral unit tests.

I think this is heading in the correct direction but I'd appreciate some feedback please.

internal/services/customization/storage_table_customization.go
```diff
@@ -0,0 +1,79 @@
...
+		bodyMap, ok := body.(map[string]interface{})
+		if !ok {
+			return nil, fmt.Errorf("expected body for %s to be an object", id.AzureResourceType)
+		}
+		for key, value := range bodyMap {
+			payload[key] = value
+		}
+	}
+
+	if rawName, ok := payload["TableName"]; ok {
```
LingyuTang, line 65: Since the TableName is fully derived from the resource `name` properties. I would suggest rejecting `TableName` input from user directly, which will keep the contract cleaner and can drop the check code. 

Also note the current key lookup is case-sensitive, so a user writing `tableName` would slip a second, service-ignored key into the payload. Consider using `strings.EqualFold` closes that gap.

internal/services/customization/storage_table_entity_customization.go
```diff
@@ -0,0 +1,119 @@
...
+
+type StorageTableEntityCustomization struct{}
+
+func (c StorageTableEntityCustomization) GetResourceType() string {
+	return "Microsoft.Storage/storageAccounts/tableServices/tables/entities"
+}
+
+func (c StorageTableEntityCustomization) CreateFunc() CreateFunc {
+	return func(ctx context.Context, client clients.Client, id parse.DataPlaneResourceId, body interface{}, options clients.RequestOptions) error {
+		payload, err := buildStorageTableEntityBody(id, body)
```
LingyuTang, line 21: This Create use insert or replace (https://learn.microsoft.com/en-us/rest/api/storageservices/insert-or-replace-entity), which means it silently overwrite if the entity already exists.

From Terraform provider convention, Create usually do a GET to check existing object then throw out error if that exists. I would suggest consider follow that as it is clearer contract and same behavior as other resource in azapi. So this should be insert https://learn.microsoft.com/en-us/rest/api/storageservices/insert-entity operation

internal/services/customization/storage_table_entity_customization.go
```diff
@@ -0,0 +1,119 @@
...
+			payload[key] = value
+		}
+	}
+
+	partitionKey, rowKey, err := storageTableEntityKeys(id)
+	if err != nil {
+		return nil, err
+	}
+
+	if rawPartitionKey, ok := payload["PartitionKey"]; ok {
```
LingyuTang, line 100: Same as the "TableName" in storage_table, might be cleaner contract just reject it outright.

LingyuTang (commented):
Hey @kewalaka, thanks for continuesly improving the PR. I think it is in a better shape and good direction. I left some comments on the Table and Table Entity customization. Please let me know if it makes sense and if you have any question. 

For the `entitiesCollection`, I can understand that it makes senses to have it for data_source of Table. But at this stage I don't feel we should support this. The reason being AzApi is designed to be thin layer of Azure API, so we generally only want to expose resources that map naturally to a real Azure API type. Since `entitiesCollection` is a synthetic type, we've introduced rather than something the Azure API defines, I'd prefer to leave it out for now.

For the `headers` and `query_parameter`, we have concerns about the security of allowing user inject headers, so we are considering moving into a whitelist implmenetation. That discussion is still early and will likely need a separate PR to address properly. Until then, we may want to limit support for Storage Table to resource only.

Let me know your thoughts. Happy to discuss further.

PR #1145, opened 2026-06-17
Acquire a policy token when a request is blocked by an invoke policy

gerrytan:
### Summary

Azure Policy *invoke policies* (a.k.a. *external evaluation* policies) require write requests to carry a **policy token** proving that the external endpoint has evaluated the operation. Without it, ARM rejects the request with `403 Forbidden` / `RequestDisallowedByPolicy`.

This PR teaches the AzAPI provider to obtain and attach that policy token automatically, so resources protected by an invoke policy can be created/updated/deleted with no special configuration. It supports two modes:

- **Reactive (default):** the provider waits for a qualifying `403`, calls `Microsoft.Authorization/acquirePolicyToken`, then rewinds and retries the original request with the token attached via the `x-ms-policy-external-evaluations` header.
- **Proactive (opt-in):** with the new `always_acquire_policy_token` attribute, the provider acquires a token up front and attaches it to every write request, avoiding the extra round-trip when a large number of resources are known to require a token. If no usable token is returned, it falls back to the reactive flow.

### Behaviour notes

- Only write methods (`PUT`/`POST`/`PATCH`/`DELETE`) are affected; `GET` is never touched.
- When no token is issued (operation denied), the original policy error is surfaced unchanged.
- Change references are not yet unsupported.

### How to enable proactive acquisition

```hcl
provider "azapi" {
  always_acquire_policy_token = true
}
```

or `export ARM_ALWAYS_ACQUIRE_POLICY_TOKEN=true`.


docs/guides/feature_acquire_policy_token.md
```diff
@@ -0,0 +1,97 @@
...
+description: |-
+  This guide explains how the AzAPI provider reactively acquires a policy token when a request is blocked by an Azure Policy invoke (external evaluation) policy, and automatically retries the request.
+
+---
+
+# Feature: Acquire Policy Token
+
+Azure Policy supports *invoke policies* (also known as *external evaluation* policies). When a resource operation is in scope of such a policy, Azure Resource Manager (ARM) calls an external endpoint to evaluate whether the operation should be allowed. To prove that the external evaluation has been performed, ARM expects the request to carry a **policy token**.
+
+When the AzAPI provider sends a write request that is in scope of one of these policies but does not yet carry a policy token, ARM rejects the request with `403 Forbidden` and a `RequestDisallowedByPolicy` error. The AzAPI provider detects this specific denial, **reactively acquires a policy token** on your behalf, and **automatically retries** the original request with the token attached.
```
Celinadhh, line 13: Should mention missingPolicyTokenDetails here for clarity

internal/clients/acquire_policy_token_policy.go
```diff
@@ -0,0 +1,212 @@
...
+		return nil, fmt.Errorf("the request was disallowed by an invoke policy that requires a change reference, but change reference is not yet supported by this provider")
+	}
+
+	log.Printf("[DEBUG] acquire policy token: %s %s was disallowed by an invoke policy that requires a policy token, attempting to acquire one", req.Raw().Method, req.Raw().URL.Path)
+	token, err := p.acquirePolicyToken(req)
+	if err != nil {
+		return nil, fmt.Errorf("the request was disallowed by an invoke policy and acquiring a policy token to satisfy it failed: %w (original policy error: %w)", err, runtime.NewResponseError(resp))
+	}
+	if token == "" {
+		log.Printf("[DEBUG] acquire policy token: no usable policy token was returned, returning the original policy denial")
```
Celinadhh, line 146: May be useful to also output the token response so user knows why no token was issued.

docs/guides/feature_acquire_policy_token.md
gerrytan: It's already mentioned below at 'How it works' point number 3.

internal/clients/acquire_policy_token_policy.go
gerrytan: Implemented: 77120b9bc

gerrytan:
Thank you @Celinadhh , I have added additional logic to detect errors on 200 responses from acquirePolicyToken API. Can you please take another look.

Also triggered acctest run: https://dev.azure.com/azclitools/internal/_build/results?buildId=323419&view=results (access restricted)

gerrytan:
@magodo can you please also review this PR from azapi maintainer perspective.

magodo:
In my opinion, this shall not be a retry-on-failure behavior. Instead, the user shall explicitly enable the policy token acquisition middleware as the policy is enabled by the user's org. A retry-on-failure behavior here is implicit and can cause unnecessary performance downgrade.

Also, the feature document (md file) is unnecessary, a simple description with a link to the Azure public document is all needed.

gerrytan:
Thanks @magodo , as discussed in our call: the performance impact will depend on the proportion of updated resources in scope of the invoke policy for the Terraform deployment. I have implemented a provider attribute to proactively acquire the token and reduce the extra round trip if the proportion & number of affected resource is known to be large beforehand. The default behaviour is reactive.

The public docs for this is not yet live, I will dedup the content in the markdown once it become available.

Can you please have another look at this PR.

Triggered acctest run: https://dev.azure.com/azclitools/internal/_build/results?buildId=323462&view=results

Celinadhh:
LGTM, thank you! 👍

internal/services/azapi_resource_acquire_policy_token_test.go
```diff
@@ -0,0 +1,322 @@
...
+type ResourceWithAcquirePolicyToken struct{}
+
+func TestAccResourceWithAcquirePolicyToken_allowedByPolicy(t *testing.T) {
+	data := acceptance.BuildTestData(t, "azapi_resource", "storage")
+	r := ResourceWithAcquirePolicyToken{}
+
+	data.ResourceTest(t, r, []resource.TestStep{
+		{
+			Config:            r.basic(data, true),
+			ExternalProviders: common.ExternalProvidersAzurermVersionFour(),
```
gerrytan, line 28: Previous run of acctests are all passing except for test in this file due to missing this AzureRM provider declaration. It is now fixed and another test rerun underway: https://dev.azure.com/azclitools/internal/_build/results?buildId=323744&view=results

There is existing method `externalProvidersAzurerm` in internal/services/azapi_data_plane_resource_upgrade_test.go but it points to AzureRM version 3. I will create a follow up PR to bump them into AzureRM version 4 and move the declaration into the common package to limit the scope of this PR diff.

gerrytan:
@magodo can you please take another look at this PR.

The two failures are suspected to be race condition on policy assignments: https://dev.azure.com/azclitools/internal/_build/results?buildId=323773&view=results

They pass after subsequent reruns: https://dev.azure.com/azclitools/internal/_build/results?buildId=323800&view=results

magodo (approved):

PR #1147, opened 2026-06-25
Require api-version query parameter when importing azapi_resource

gerrytan:
## Summary

During import, the provider cannot determine the target api-version of a resource. When the `api-version` query parameter is omitted (ID-based import) or the `type` is omitted (identity-based import), the provider falls back to the latest [indexed](https://github.com/Azure/terraform-provider-azapi/blob/main/internal/azure/generated/index.json) api-version (regardless of preview / stable), which can silently cause an api-version mismatch.

This PR makes the `api-version` query parameter expected during import:

- In **2.x**, a missing `api-version` emits a **warning** and keeps the existing fall-back behaviour (backwards compatible).
- In **3.0** (behind the `ARM_THREEPOINTZERO_BETA` feature flag), a missing `api-version` becomes a hard **error**.
- A new `disable_mandatory_import_api_version` provider option (also sourced from the `ARM_DISABLE_MANDATORY_IMPORT_API_VERSION` environment variable) lets users opt out of the 3.0 error and revert to the 2.x warning/fall-back behaviour.

## Notes for reviewers

- The `disable_mandatory_import_api_version` escape hatch preserves the 2.x behaviour even under 3.0. This is implemented to ease the 3.0 migration effort for users that don't see any adverse impact


.github/workflows/unit-test.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
       - run: go generate ./...
       - name: git diff
         run: |
-          git diff --exit-code || \
+          git diff --exit-code --ignore-blank-lines --ignore-space-at-eol || \
```
gerrytan, line 36: This is needed to avoid perpetual diff, the doc generator and the IDE / git whitespace EOL configuration doesn't always agree

magodo (dismissed):

magodo:
The issue this PR is trying to solve is actually a built-in API type issue, which instead of enforcing the user to specify the api vesion (which downgrades the UX), I'd prefer https://github.com/Azure/terraform-provider-azapi/pull/1093.

gerrytan:
Thanks @magodo, summary after discussion on our calls: the breaking change impact of forcing api-version is deemed to significant for users. Instead we should optimise the automatic API version fallback logic, and make sure the logging, error message and docs are clear enough to support this.

PR #1150, opened 2026-07-01
Support sensitive body in resource action

### LINKED ISSUE #1039: feature request: support ephemeral values on `azapi_resource_action` action and resource
### Use Case
Managing ephemeral secrets (like VPN pre-shared keys etc.) that should:
- Never be stored in Terraform state
- Only be updated when explicitly triggered via lifecycle events
- Use ephemeral resources for generation

### Current Limitation
```terraform
action "azapi_resource_action" "preshared_key_put" {
  config {
    type        = "Microsoft.Network/vpnGateways/vpnConnections/vpnLinkConnections/sharedKeys@2025-03-01"
    resource_id = "${local.vpnLinkKeyId}/sharedKeys/default"
    method      = "PUT"

    body = {
      properties = {}
    }

    # ❌ Not supported
    sensitive_body = {
      properties = {
        sharedKey = ephemeral.random_password.link_connection_shared_key[local.vpnLinkKeyId].result
      }
    }
  }

resource "azapi_resource_action" "preshared_key_put" {
  config {
    type        = "Microsoft.Network/vpnGateways/vpnConnections/vpnLinkConnections/sharedKeys@2025-03-01"
    resource_id = "${local.vpnLinkKeyId}/sharedKeys/default"
    method      = "PUT"

    body = {
      properties = {}
    }

    # ❌ Not supported
    sensitive_body = {
      properties = {
        sharedKey = ephemeral.random_password.link_connection_shared_key[local.vpnLinkKeyId].result
      }
    }
    # ❌ Not supported
    sensitive_body_version = {
      "properties.sharedKey" = terraform_data.link_connection_shared_key_version[local.vpnLinkKeyId].output.key_version
    }
  }
}

 # ❌ Renews (and writes new, ephemeral secret) on each apply, not only when action is required)
ephemeral "azapi_resource_action" "preshared_key_put" {

  type        = "Microsoft.Network/vpnGateways/vpnConnections/vpnLinkConnections/sharedKeys@2025-03-01"
  resource_id = "${local.vpnLinkKeyId}/sharedKeys/default"
  method      = "PUT"

  body = {
    properties = {
      sharedKey =  ephemeral.random_password.link_connection_shared_key[local.vpnLinkKeyId].result
    }
  }
}
```

### Workaround Required
Currently must use `azapi_update_resource` which:
- Creates a managed resource in state (even though data is encrypted)
- Is less intuitive for one-time operations
- Doesn't align with the action pattern
- Does not support other operations than 'PATCH' (`method`)

Using the _ephemeral_ `azapi_resource_action` does not serve the purpose as it will update on every apply.

### Proposed Solution
Add `sensitive_body` support to `action` blocks, mirroring the existing implementation in `azapi_resource` and `azapi_update_resource`. Action probably does not need to support `sensitive_body_version` as the trigger logic should really happen in the resource calling the action.

Consider adding both, `sensitive_body` and `sensitive_body_version` to the _resource_ `azapi_update_resource` and maybe the _ephemeral_ `azapi_update_resource`. This would allow similar, secure patterns for users that cannot yet make use of an action block.

### Benefits
1. True ephemeral secret management
2. Better alignment with Terraform 1.10+ patterns
3. Cleaner separation between managed resources and operations
4. Consistency across AzAPI provider resources

### References
- `azapi_resource` already supports `sensitive_body`
- `azapi_update_resource` already supports `sensitive_body`
- Action blocks support lifecycle triggers
- [988](https://github.com/Azure/terraform-provider-azapi/issues/988) - which I believe could have been closed as the action support was released with [v2.8.0](https://github.com/Azure/terraform-provider-azapi/blob/main/CHANGELOG.md#v280)

hqhqhqhqhqhqhqhqhqhqhq:
## Description

Adds `sensitive_body` support to `azapi_resource_action` across the action, managed resource, and ephemeral resource forms.

For the managed resource, `sensitive_body` is write-only and is merge-patched into `body` before sending the request. `sensitive_body_version` follows the existing `azapi_resource` and `azapi_update_resource` pattern so callers can control when individual write-only fields are resent. The resource schema is upgraded to v3 with a state migration for existing v2 state.

For Terraform actions, `sensitive_body` is a write-only action argument so ephemeral secrets can be passed to the action without being serialized into the saved plan. For ephemeral resources, `sensitive_body` is merged into the request body while relying on the ephemeral resource lifecycle for non-persistence.

`sensitive_body` is rejected with `when = "destroy"` on the managed resource because write-only configuration is not available during destroy.

Fixes #1039
Related to #1130

### DIFF
--- a/docs/actions/resource_action.md
+++ b/docs/actions/resource_action.md
@@ -148,4 +148,4 @@ action "azapi_resource_action" "power_on" {
 
 	-> Value must be one of: ["POST" "PATCH" "PUT" "DELETE" "GET" "HEAD"].
 - `query_parameters` (Map of List of String) Query parameters to include in the request.
-- `sensitive_body` (Dynamic, [Write-only](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments)) A dynamic attribute that contains the write-only properties of the request body. This will be merge-patched to the body to construct the actual request body.
+- `sensitive_body` (Dynamic, [Write-only](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments)) A dynamic attribute that contains the write-only properties of the request body. This will be merge-patched to the body to construct the actual request body. If a property is defined in both `body` and `sensitive_body`, the `sensitive_body` value takes precedence.
--- a/docs/actions/resource_action.md
+++ b/docs/actions/resource_action.md
@@ -148,3 +148,4 @@ action "azapi_resource_action" "power_on" {
 
 	-> Value must be one of: ["POST" "PATCH" "PUT" "DELETE" "GET" "HEAD"].
 - `query_parameters` (Map of List of String) Query parameters to include in the request.
+- `sensitive_body` (Dynamic, [Write-only](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments)) A dynamic attribute that contains the write-only properties of the request body. This will be merge-patched to the body to construct the actual request body.
--- a/internal/services/azapi_resource_action_ephemeral.go
+++ b/internal/services/azapi_resource_action_ephemeral.go
@@ -177,7 +177,7 @@ func (r *ActionEphemeral) Open(ctx context.Context, request ephemeral.OpenReques
 
 	ctx = tflog.SetField(ctx, "resource_id", id.ID())
 
-	requestBody, err := buildActionRequestBody(model.Body, model.SensitiveBody, types.MapNull(types.StringType), types.MapNull(types.StringType))
+	requestBody, err := buildActionEphemeralRequestBody(model)
 	if err != nil {
 		response.Diagnostics.AddError("Invalid request body", err.Error())
 		return
@@ -221,6 +221,10 @@ func (r *ActionEphemeral) Open(ctx context.Context, request ephemeral.OpenReques
 	model.Output = output
 
 	response.Diagnostics.Append(response.Result.Set(ctx, model)...)
+}
+
+func buildActionEphemeralRequestBody(model ActionEphemeralModel) (interface{}, error) {
+	return buildActionRequestBody(model.Body, model.SensitiveBody, types.MapNull(types.StringType), types.MapNull(types.StringType))
 }
 
 func (r *ActionEphemeral) RenderOption() tffwdocs.EphemeralResourceRenderOption {
--- a/internal/services/azapi_resource_action_ephemeral.go
+++ b/internal/services/azapi_resource_action_ephemeral.go
@@ -32,6 +32,7 @@ type ActionEphemeralModel struct {
 	Action               types.String     `tfsdk:"action"`
 	Method               types.String     `tfsdk:"method"`
 	Body                 types.Dynamic    `tfsdk:"body"`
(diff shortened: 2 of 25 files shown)


docs/actions/resource_action.md
```diff
@@ -148,3 +148,4 @@ action "azapi_resource_action" "power_on" {
 
 	-> Value must be one of: ["POST" "PATCH" "PUT" "DELETE" "GET" "HEAD"].
 - `query_parameters` (Map of List of String) Query parameters to include in the request.
+- `sensitive_body` (Dynamic, [Write-only](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments)) A dynamic attribute that contains the write-only properties of the request body. This will be merge-patched to the body to construct the actual request body.
```
gerrytan: @hqhqhqhqhqhqhqhqhqhqhq what is the behaviour if `sensitive_body` contains the same key as `body`? Perhaps we should document this and/or invalidate this?

Some concern applies to other resources / types.

internal/services/azapi_resource_action_ephemeral.go
```diff
@@ -171,9 +177,9 @@ func (r *ActionEphemeral) Open(ctx context.Context, request ephemeral.OpenReques
 
 	ctx = tflog.SetField(ctx, "resource_id", id.ID())
 
-	var requestBody interface{}
-	if err := unmarshalBody(model.Body, &requestBody); err != nil {
-		response.Diagnostics.AddError("Invalid body", fmt.Sprintf(`The argument "body" is invalid: %s`, err.Error()))
+	requestBody, err := buildActionRequestBody(model.Body, model.SensitiveBody, types.MapNull(types.StringType), types.MapNull(types.StringType))
+	if err != nil {
+		response.Diagnostics.AddError("Invalid request body", err.Error())
```
gerrytan, line 182: Can you please also provide test coverage for `azapi_resource_action` and `azapi_resource_action_ephemeral`.

gerrytan:
Please have a look at internal/clients/live_traffic_log_policy.go, I think it will log the sensitive body (including existing sensitive body).

I think we should also alter the behaviour of this to not log sensitive content at all.

gerrytan (changes requested):
Thanks @hqhqhqhqhqhqhqhqhqhqhq , I left few questions / suggestions.

Also triggered the acctests for this PR: https://dev.azure.com/azclitools/internal/_build/results?buildId=326940. Note that branch with `with-acctest/` prefix will auto-trigger that pipeline.

gerrytan (changes requested):
Thank you for addressing the feedbacks @hqhqhqhqhqhqhqhqhqhqhq . Just 1 last thing:

Can you please take a look at the acctest failure here: https://dev.azure.com/azclitools/b5336d23-77d8-49b8-9327-6c483cc524dd/_apis/build/builds/326940/logs/15

`--- FAIL: TestAccAzapiActionResourceUpgrade_` (and other failing tests)

They may be related to changes in this PR?

gerrytan (approved):

ziyeqf:
thou the PR has been merged after a long while, but it breaks armstrong a bit... armstrong depends on azapi's request body to caculate coverage percentage

PR #1151, opened 2026-07-02
fix: Support PFX modern encryption for client_certificate auth

### LINKED ISSUE #1052: PFX modern encryption not supported
When using the AZAPI Terraform provider with client_certificate and client_certificate_password it seems only pfx in legacy encryption (MAC: sha1, PKCS7 Encrypted data: pbeWithSHA1And40BitRC2-CBC) are supported but not modern PFX encryption (MAC: sha256,
PKCS7 Encrypted data: PBES2, PBKDF2, AES-256-CBC, Iteration 2048, PRF hmacWithSHA256).

This is even more confusing since the "azurerm" provider and the "azuread" provider both understand the new encryption correctly while configured in the exact same way as the azapi provider.

Example: "azapi" fails with the following config while "azurerm" works fine:
```
provider "azapi" {
  use_cli            = false
  tenant_id          = var.azure_tenant_id
  subscription_id    = var.azure_subscription_id
  client_id          = var.client_id
  client_certificate = data.azurerm_key_vault_secret.service_principal1_pfx_base64.value
  client_certificate_password = data.azurerm_key_vault_secret.service_principal1_pfx_pw.value
}
```

```
provider "azurerm" {
  tenant_id          = var.azure_tenant_id
  subscription_id    = var.azure_subscription_id
  client_id          = var.client_id
  client_certificate = data.azurerm_key_vault_secret.service_principal1_pfx_base64.value
  client_certificate_password = data.azurerm_key_vault_secret.service_principal1_pfx_pw.value
}
```

Error message from azapi in Terrafrom debug logs:
`DEBUG] failed to build client certificate cred option: failed to parse client certificate": pkcs12: unknown digest algorithm: 2.16.840.1.101.3.4.2.1`

gerrytan:
Fixes #1052

### Problem

When authenticating with `client_certificate` + `client_certificate_password`, AzAPI only accepted PFX bundles using **legacy** encryption (MAC: SHA1, PKCS7 `pbeWithSHA1And40BitRC2-CBC`). **Modern** PFX bundles (MAC: SHA256, PKCS7 PBES2 / PBKDF2 / AES-256-CBC) failed with:

```
failed to parse client certificate": pkcs12: unknown digest algorithm: 2.16.840.1.101.3.4.2.1
```

The root cause is an upstream limitation in `azidentity.ParseCertificates`, and [azure-sdk-for-go#22906](https://github.com/Azure/azure-sdk-for-go/issues/22906) confirmed modern PFX will **not** be supported there.

### Solution

The fix lands in [`Azure/entrauth`](https://github.com/Azure/entrauth) ([PR #4](https://github.com/Azure/entrauth/pull/4)) and is consumed here by bumping the dependency. When `azidentity.ParseCertificates` fails, `getClientCert` now falls back to the `software.sslmate.com/src/go-pkcs12` parser (`pkcs12.DecodeChain`), which understands modern PFX encryption. Legacy PFX continues to parse via the existing `azidentity` path, so behavior is unchanged for existing users.

Changes:
- Bump `github.com/Azure/entrauth` to the version containing the pkcs12 fallback.

### Testing

This code path can't be exercised by the standard acceptance tests without a real certificate, so I verified it manually:

1. Produced **both a legacy and a modern PFX** certificate following the approach in `aztfauth/aztfauth_test.go` (`TestGetClientCert_ParsesLegacyAndModernPFX`), which generates a self-signed cert/key and exports two PFX files via OpenSSL:
   - **Legacy:** `openssl pkcs12 -export -certpbe PBE-SHA1-3DES -keypbe PBE-SHA1-3DES -macalg sha1 ...`
   - **Modern:** `openssl pkcs12 -export -macalg sha256 ...`
2. Set `ARM_CLIENT_CERTIFICATE_PATH` (and the certificate password) to each generated PFX in turn and ran the acceptance test `TestAccAuth_clientCertificatePath`.

I also ran regression acctest: https://dev.azure.com/azclitools/internal/_build/results?buildId=327110


gerrytan:
Hi @magodo, can you please take a look at this PR to propagate https://github.com/Azure/entrauth/pull/4 into AzAPI.

magodo (approved):

PR #1156, opened 2026-07-08
Fix missing resource identity after read

### LINKED ISSUE #1050: Microsoft.Resources/deploymentScripts@2023-08-01 - Missing Resource Identity After Read during terraform plan
After updating to version 2.8.0 of azapi (with Terraform 1.14.4) `terraform plan` fails with the following error:

```
Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: Missing Resource Identity After Read
│ 
│   with azapi_resource.kv_01_initialize_secrets[0],
│   on kc_kv_01_secrets.tf line 8, in resource "azapi_resource" "kv_01_initialize_secrets":
│    8: resource "azapi_resource" "kv_01_initialize_secrets" {
│ 
│ The Terraform Provider unexpectedly returned no resource identity data after having no errors in the resource read. This is always an issue in the Terraform Provider and should be reported to the provider developers.
╵
```
The problem happens with a deploymentScripts Resource:
```hcl
resource "azapi_resource" "kv_01_initialize_secrets" {
  type      = "Microsoft.Resources/deploymentScripts@2023-08-01"
  name      = "kv_secret"
  parent_id = azapi_resource.kv_rg.id
  location  = azapi_resource.kv_rg.location

  schema_validation_enabled = false

  body = {
    kind = "AzurePowerShell"
    properties = {
      azPowerShellVersion = "14.0"
      retentionInterval   = "PT1H"
      cleanupPreference   = "Always"
      timeout             = "PT30M"
      arguments           = "-KeyVaultName '${azapi_resource.kv.name}' -SecretNames "somesecret"
      scriptContent       = <<EOF
        "powershell script to add secret to kv"
EOF
    }
  }
}
````
The special thing about the deploymentScripts Resources is that it has an expiration date and is automatically deleted afterward. https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/deployment-script-develop?tabs=CLI#clean-up-deployment-script-resources
As long as the expiration date is not reached the plan still works. But after the resource was deleted the error accours. Interestingly this only happens with provider version 2.8.0 (with 2.7.0 the plan still works even when the deploymentScripts resource was cleaned up/deleted). 

The unique aspect of `deploymentScripts` resources is that they have an expiration date and are automatically deleted thereafter. As long as the expiration date is not reached, the plan continues to function. However, after the resource has been deleted, the error occurs. Interestingly, this issue starts with provider version 2.8.0. With version 2.7.0, the plan continues to work even after the deploymentScripts resource has been cleaned up or deleted.

### PR DESCRIPTION
**Pull request by hqhqhqhqhqhqhqhqhqhqhq**
## Summary
- Populate `azapi_resource` identity from the existing state before reading the Azure resource.
- Add a focused 404 unit regression.
- Add an upgrade acceptance regression that creates state with AzAPI v2.7.0, deletes the resource outside Terraform, and plans with the current provider.

## Root Cause
AzAPI v2.7.0 did not store Terraform resource identity. After upgrading, a successful read populated identity at the end of `Read()`, but a 404 removed the resource from state and returned before that happened. Terraform Plugin Framework then reported `Missing Resource Identity After Read`.

Fresh resources created by current provider versions already have identity in state, which is why deleting a newly created v2.10.0 resource does not reproduce the issue.

Fixes #1050

### DIFF
--- a/internal/services/azapi_resource.go
+++ b/internal/services/azapi_resource.go
@@ -1006,6 +1006,9 @@ func (r *AzapiResource) Read(ctx context.Context, request resource.ReadRequest, 
 	if response.Diagnostics.Append(request.State.Get(ctx, &model)...); response.Diagnostics.HasError() {
 		return
 	}
+	if response.Diagnostics.Append(response.Identity.SetAttribute(ctx, path.Root("id"), model.ID)...); response.Diagnostics.HasError() {
+		return
+	}
 
 	readTimeout, diags := model.Timeouts.Read(ctx, 5*time.Minute)
 	response.Diagnostics.Append(diags...)
@@ -1159,7 +1162,6 @@ func (r *AzapiResource) Read(ctx context.Context, request resource.ReadRequest, 
 	}
 
 	response.Diagnostics.Append(response.State.Set(ctx, state)...)
-	response.Diagnostics.Append(response.Identity.SetAttribute(ctx, path.Root("id"), state.ID)...)
 }
 
 func (r *AzapiResource) Delete(ctx context.Context, request resource.DeleteRequest, response *resource.DeleteResponse) {
--- a/internal/services/azapi_resource_identity_test.go
+++ b/internal/services/azapi_resource_identity_test.go
@@ -21,7 +21,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
 )
 
-func TestAzapiResourceReadSetsIdentityWhenArmReturnsEmptyObject(t *testing.T) {
+func TestAzapiResourceReadSetsIdentityWhenArmReturnsNotFound(t *testing.T) {
 	ctx := context.Background()
 	resourceID := "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/acctest-rg"
 	resourceType := "Microsoft.Resources/resourceGroups@2021-04-01"
@@ -29,7 +29,7 @@ func TestAzapiResourceReadSetsIdentityWhenArmReturnsEmptyObject(t *testing.T) {
 	resourceClient, err := clients.NewResourceClient(staticTokenCredential{}, &arm.ClientOptions{
 		ClientOptions: policy.ClientOptions{
 			Transport: staticTransport{
-				statusCode: http.StatusOK,
+				statusCode: http.StatusNotFound,
 				body:       `{}`,
 			},
 		},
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: CHANGELOG.md [hunk 1]
```diff
@@ -10,6 +10,7 @@ ENHANCEMENTS:
 
 BUG FIXES:
 
+- Fix `Missing Resource Identity After Read` error when an `azapi_resource` is deleted outside Terraform and a subsequent plan removes it from state (GH-1023).
```
gerrytan: We cannot add a changelog entry retrospectively, v2.10.0 is already released. Let's just do a batch changelog update later when we're making another release.

internal/services/azapi_resource.go
```diff
@@ -1036,6 +1036,16 @@ func (r *AzapiResource) Read(ctx context.Context, request resource.ReadRequest,
...
+			if request.Identity != nil && !request.Identity.Raw.IsNull() {
+				var resourceIdentity AzapiResourceIdentityModel
+				response.Diagnostics.Append(request.Identity.Get(ctx, &resourceIdentity)...)
+				if response.Diagnostics.HasError() {
+					return
+				}
+				response.Diagnostics.Append(response.Identity.Set(ctx, resourceIdentity)...)
+			} else {
+				response.Diagnostics.Append(response.Identity.SetAttribute(ctx, path.Root("id"), model.ID)...)
+			}
```
gerrytan: @hqhqhqhqhqhqhqhqhqhqhq I just tried running the acctest you added with this section commented out and it still passes. I don't think this is required. Can you please double-check.

gerrytan (changes requested):

CHANGELOG.md
hqhqhqhqhqhqhqhqhqhqhq: Removed the retrospective changelog entry and corrected the PR description so it no longer claims one.

internal/services/azapi_resource.go
hqhqhqhqhqhqhqhqhqhqhq: I reproduced the missing-identity error with the upgrade path that the earlier test missed: create the resource with AzAPI v2.7.0, delete it outside Terraform, then plan with the current provider on Terraform 1.14.4.

Without the Read change, that test fails with `Missing Resource Identity After Read`; with the change, it passes. A fresh resource created by v2.10.0 does not reproduce the issue because identity is already stored in state.

I replaced the previous acceptance scenario with this red/green upgrade regression and moved identity population to the start of `Read()` so the 404 path returns it before removing the resource from state.

gerrytan (approved):
Thank you @hqhqhqhqhqhqhqhqhqhqhq , this change looks good to me

PR #1158, opened 2026-07-09
`azapi_resource_action` - fix unexpected value error of `exist` when updating existing actions

### LINKED ISSUE #1094: Logic bug in `azapi_resource_action.exist` for resources created before v2.9.0
The addition of the read-only `exist` field for the `azapi_resource_action` resource has resulted in an error when a resource created before the v2.9.0 release gets modified.

> ╷
> │ Error: Provider produced inconsistent result after apply
> │ 
> │ When applying changes to
> │ module.aks.module.additional_cluster_tags_v1.azapi_resource_action.default,
> │ provider "provider[\"registry.terraform.io/azure/azapi\"]" produced an
> │ unexpected new value: .exist: was null, but now cty.True.
> │ 
> │ This is a bug in the provider, which should be reported in the provider's
> │ own issue tracker.
> ╵

This is caused because #1029 didn't update the schema or add a plan modifier to support setting `exist` as part of the plan. In #1057 a plan modifier was added, but this doesn't function correctly.

It looks to me like the fix for this is to remove the `boolplanmodifier.UseStateForUnknown()` for `exist`, as this returns `null` for a resource created before `v2.9.0` (and if `false` is ever persisted would also break for a valid apply), and manage `exist` in `ModifyPlan()`. Alternatively you could do a state migration to add `exist`, but you'd still need to deal with planning the value of `exist`.

CC @ms-henglu @magodo

LingyuTang:
### Description
This PR fixes a bug where updating an existing `azapi_resource_action` could fail due to the computed `exist` field.

The root cause was `boolplanmodifier.UseStateForUnknown()`, which could propagate a prior null value for `exist` into the plan. During apply, the provider returns a concrete value (for example, true), which Terraform then reports as an unexpected new value.

Changes
1. Use the prior state value as the default for exist during planning.
2. Set exist to unknown only when the action is expected to run during apply.

acctest: https://dev.azure.com/azclitools/internal/_build/results?buildId=328603&view=results
fix https://github.com/Azure/terraform-provider-azapi/issues/1094


gerrytan (approved):
Thanks @LingyuTang , I've tested the change manually, upgrading from 2.8.0, and the problem stated in the issue is now solved.

While I don't quite like the original design of "exist" and "ignore_not_found" attributes, I guess we're stuck with it now.

This PR is good to merge, however I note that the acc test coveraged were skipped in the past due to quota. Can you please raise a follow up PR to refactor the test to use a storage account and POST /listKeys action (or other that's not bound by quota) and re-enable them.

PR #1160, opened 2026-07-10
fix: Return initial PUT body from ResourceClient CreateOrUpdate

wuxu92:
The CreateOrUpdate method should return the origin PUT response, not the polling response. it's required for some special resources as they depends on some specific property value of the PUT response, like role definition.


gerrytan:
Merged latest main to apply workflow fix, ran acctest pipeline: https://github.com/Azure/terraform-provider-azapi/pull/1160/checks?check_run_id=86301148102

gerrytan:
This PR looks good, thx @wuxu92 for the contrib. The acctest failure was unrelated to this change (quota issue).

However currently there's unrelated problem in main that will be fixed in #1166. We'll have to wait until that merge first and then refresh this.

gerrytan:
Merged main to fix the generated doc drift. The unit test workflow is now passing.

Merging this PR.

gerrytan (approved):

PR #1164, opened 2026-07-10
Fix: OIDC token file path logic

### LINKED ISSUE #1149: use_aks_workload_identity = false still picks up AZURE_FEDERATED_TOKEN_FILE (breaks GitHub OIDC on AKS-hosted runners, since 2.6.0)
We run Terraform on self-hosted GitHub Actions runners hosted in AKS, authenticating to azapi with GitHub OIDC (`use_oidc = true`). The AKS pods have Workload Identity enabled, so `AZURE_FEDERATED_TOKEN_FILE` (plus `AZURE_CLIENT_ID`/`AZURE_TENANT_ID`) is injected into the environment for the node identity.

Even with `use_aks_workload_identity` left at its default (`false`), azapi picks up that `AZURE_FEDERATED_TOKEN_FILE` and tries the workload-identity token instead of our GitHub OIDC credential, failing with `AADSTS700211`:

```hcl
provider "azapi" {
  use_oidc = true
  # use_aks_workload_identity defaults to false
}
```
```
WorkloadIdentityCredential authentication failed.
AADSTS700211: No matching federated identity record found for the presented assertion.
```

The `OIDCTokenFilePath` default in `internal/provider/provider.go` isn't gated by `use_aks_workload_identity` (unlike the `AZURE_TENANT_ID` branch just above it, which is):

```go
} else if v := os.Getenv("AZURE_FEDERATED_TOKEN_FILE"); v != "" {
    model.OIDCTokenFilePath = types.StringValue(v)   // picked up even when use_aks_workload_identity = false
}
```

Because `UseOIDCTokenFile = UseOIDC || UseAKSWorkloadIdentity`, the file credential lands ahead of the GitHub one in the chain, and since it fails with a hard rejection rather than reporting "unavailable", the chain never falls through to GitHub.

This started on 2.6.0 (2.5.0 is fine), so it looks related to the `aztfauth` switch — same area as #962 / #971. Gating the `AZURE_FEDERATED_TOKEN_FILE` fallback behind `use_aks_workload_identity` (consistent with `AZURE_TENANT_ID`) would fix it. Current workaround is to `unset AZURE_FEDERATED_TOKEN_FILE` before running Terraform.

gerrytan:
Fixes #1149 by ensuring the provider only uses AZURE_FEDERATED_TOKEN_FILE when AKS workload identity is explicitly enabled.

When use_aks_workload_identity is left at its default `false`, the provider should not pick up the AKS-injected federated token file and override GitHub OIDC authentication. This change keeps the existing OIDC behavior intact for normal OIDC flows while preventing the AKS workload identity token from hijacking the credential chain.

## Testing

- acctest: https://dev.azure.com/azclitools/internal/_build/results?buildId=328859&view=results

### DIFF
--- a/internal/provider/provider.go
+++ b/internal/provider/provider.go
@@ -96,6 +96,18 @@ type providerEndpointData struct {
 	ResourceManagerAudience      types.String `tfsdk:"resource_manager_audience"`
 }
 
+func getOIDCTokenFilePathFromEnv(useAKSWorkloadIdentity bool) string {
+	if v := os.Getenv("ARM_OIDC_TOKEN_FILE_PATH"); v != "" {
+		return v
+	}
+	if useAKSWorkloadIdentity {
+		if v := os.Getenv("AZURE_FEDERATED_TOKEN_FILE"); v != "" {
+			return v
+		}
+	}
+	return ""
+}
+
 func (p Provider) Metadata(ctx context.Context, request provider.MetadataRequest, response *provider.MetadataResponse) {
 	response.TypeName = "azapi"
 }
@@ -483,9 +495,7 @@ func (p Provider) Configure(ctx context.Context, request provider.ConfigureReque
 	}
 
 	if model.OIDCTokenFilePath.IsNull() {
-		if v := os.Getenv("ARM_OIDC_TOKEN_FILE_PATH"); v != "" {
-			model.OIDCTokenFilePath = types.StringValue(v)
-		} else if v := os.Getenv("AZURE_FEDERATED_TOKEN_FILE"); v != "" {
+		if v := getOIDCTokenFilePathFromEnv(model.UseAKSWorkloadIdentity.ValueBool()); v != "" {
 			model.OIDCTokenFilePath = types.StringValue(v)
 		}
 	}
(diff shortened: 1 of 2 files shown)


gerrytan:
The two failing acctests are flakes, it passes locally after rerun:
1. TestAccGenericResource_listUniqueIdProperty 
2. TestAccResourceWithAcquirePolicyToken_disallowedByPolicy

gerrytan:
@hqhqhqhqhqhqhqhqhqhqhq can you please review this PR

gerrytan:
@LingyuTang can you please review this PR

LingyuTang (dismissed):
Hey Gerry, I think this looks good overall. 

The only thing I'm not sure about is whether we need a separate `getOIDCTokenFilePathFromEnv` function since it's only used once. Similarly, I'm wondering if the unit test is necessary here, as it might be simpler to cover this through changing an existing acceptance test instead. But neither of these is a blocker, so happy to go with whatever you think.

gerrytan:
Thx @LingyuTang , as discussed on DM, our acctest pipeline is based on password secret, and does not yet support OIDC / federated token. I've pushed an explanation comment on the code. Can you please take another look.

LingyuTang (approved):
Looks good! Thanks for explaining the logic and adding the comment.

PR #1165, opened 2026-07-10
`azapi_resource_action` - use storage account for acc test

LingyuTang:
### Description
This PR is to: 
1. Use the storage account for most of the acc test of azapi_resource_action to avoid quotas issue
2. Remove the skip test function from those test

ACC test: https://dev.azure.com/azclitools/internal/_build/results?buildId=328898&view=results


gerrytan (approved):
Nice work @LingyuTang ! Good to see those tests are running again. This PR looks good to me

PR #1166, opened 2026-07-10
fix: doc drift on docs/index.md and update worflow path condition

gerrytan:
Fixes failing unit-test workflow due to drift between checked in markdown and generated ones.

I also fixed the path on the GH workflow file so such drift is caught in the future.

## Testing

Relying only on unit test and other GH workflow since there's no func change


LingyuTang (approved):
Thanks for the PR! LGTM.

PR #1168, opened 2026-07-14
`azapi_resource_action` - set output as unknow when action executes

LingyuTang:
### Description

Resource action outputs may change after an action executes, which can cause inconsistent post-apply state and test failures. This case the intermittent error in nightly pipeline. 

This PR fixes that behavior by marking output as unknown during planning whenever the action is expected to execute, so Terraform does not lock in a stale output type before applying.

acctest: https://dev.azure.com/azclitools/internal/_build/results?buildId=329472&view=results


gerrytan (approved):
Thanks for the fix and running acctest. This PR looks good to me @LingyuTang

PR #1170, opened 2026-07-14
azapi_update_resource: read_override property for API that require POST for fetches (GH-1167)

### LINKED ISSUE #1167: Continuous drift updating flex consumption app config
The following AzApi updates the app settings for a flex consumption app. However, terraform sees continual drift.

````
resource "azapi_update_resource" "myfunc_app_settings" {
  type      = "Microsoft.Web/sites/config@2025-03-01"
  parent_id = azapi_resource.myfunc_fca.id
  name      = "appsettings"

  body = {
    properties = {
      DEPLOYMENT_STORAGE_CONNECTION_STRING  = module.myfunc.storage_account_sensitive.primary_connection_string
      APPLICATIONINSIGHTS_CONNECTION_STRING = azurerm_application_insights.myfunc.connection_string

      AzureWebJobsStorage__credential     = "managedidentity"
      AzureWebJobsStorage__blobServiceUri = trimsuffix(module.myfunc.storage_account_sensitive.primary_blob_endpoint, "/")

      NAMESPACE   = "XXX"
    }
  }
  ignore_casing             = false
  ignore_missing_property   = false
}
````
However, terraform sees continual drift.

````
  ~ resource "azapi_update_resource" "tfprivreg_app_settings" {
      ~ body                    = {
          ~ properties = {
              ~ APPLICATIONINSIGHTS_CONNECTION_STRING = (sensitive value)
              ~ AzureWebJobsStorage__blobServiceUri   = (sensitive value)
              ~ AzureWebJobsStorage__credential       = null -> "managedidentity"
              ~ DEPLOYMENT_STORAGE_CONNECTION_STRING  = (sensitive value)
              ~ LOG_LEVEL                             = null -> "info"
              ~ NAMESPACE                             = null -> "XXX"
              ~ TEST                                  = null -> "helloword"
            }
````

Since the AzureRM has so many issues with flex consumption apps I'd rather hoped AzAPI would be easier. It isn't!

I note that the documentation says to use azapi_resource (https://learn.microsoft.com/en-us/azure/templates/microsoft.web/sites/config/appsettings?pivots=deployment-language-terraform) I tried this too. It requires a terraform import because the resource already exists and then exhibits exactly the same behaviour.

### PR DESCRIPTION
**Pull request by jaredfholgate**
## What

Adds optional `read_override` so property values can be read from a POST /list action (or other configured path) when a plain `GET` returns them empty.

Example:

```terraform
resource "azapi_update_resource" "test" {
  type      = "Microsoft.Web/sites/config@2023-12-01"
  name      = "appsettings"
  parent_id = azapi_resource.site.id
  body = {
    properties = {
      KEY1 = "value1"
      KEY2 = "value2"
      KEY3 = "value3"
    }
  }
  ignore_casing           = false
  ignore_missing_property = false
  read_override = {
    method = "POST"
    action = "list"
  }
}
```

## Why

Fixes #1167. A `GET` on `Microsoft.Web/sites/config/appsettings` returns `properties: {}` — the real values are only exposed via `POST .../config/appsettings/list`. azapi's read-merge is config-body-driven, so an empty `GET` drops every configured key and reports perpetual drift (notably for Flex Consumption function apps).

## Testing

- Unit & acctest


internal/services/azapi_update_resource.go
```diff
@@ -284,6 +288,24 @@ func (r *AzapiUpdateResource) Schema(ctx context.Context, request resource.Schem
...
+				MarkdownDescription: "The HTTP method used when `read_action` names an action to call. Defaults to `POST`, which matches the Azure convention for `list*` functions. The embedded resource metadata does not carry the verb, so set this when an action must be read with a different method.",
+				Validators: []validator.String{
+					stringvalidator.OneOf("POST", "PATCH", "PUT", "DELETE", "GET", "HEAD"),
+				},
+			},
+
+			"read_action_response_path": schema.StringAttribute{
+				Optional:            true,
+				MarkdownDescription: "An optional [JMESPath](https://jmespath.org/) expression applied to the `read_action` response before it is merged into `body`. Use this when the list function nests the values under a sub-object.",
+			},
```
gerrytan: @jaredfholgate designing 3 separate optional props like this opens possibility of supplying only 1 of them (eg: read_action_response_path without read_action_method, and it's unclear whether the POST override will be effective or not).

What do you think about designing them as a block like this:

```hcl
read_override { 
  method = "POST" // required, POST is the only valid value
  action = "list" // required
  response_path = "..." // optional
}

**Review by gerrytan — CHANGES_REQUESTED** [review]
Thanks @jaredfholgate , I can reproduce the problem reported in #1167 and I'm onboard with the idea of introducing this read override. However I have some design suggestions. Keen to hear your thoughts.

File: internal/services/azapi_update_resource.go (hunk 1 above)
Comment (jaredfholgate) [reply internal/services/azapi_update_resource.go]: @gerrytan  Blocks are deprecated and not recommended in the latest SDK. It would need to be an object instead, like this:

```hcl
read_override = { 
  method        = "POST" // required, POST is the only valid value
  action        = "list" // required
  response_path = "..." // optional
}
```

I'm fine with updating it to be an object, I think I was just following the existing convention like headers where everything seems to be separate inputs. Makes sense to group them though.

jaredfholgate: Link with the recommendation to use nested types instead of blocks: https://developer.hashicorp.com/terraform/plugin/framework/handling-data/blocks

gerrytan: Nice, TIL. Thx @jaredfholgate , I'll push some modification to this branch per discussion above.

internal/services/read_action.go
```diff
@@ -0,0 +1,163 @@
...
+	diags := value.As(ctx, &model, basetypes.ObjectAsOptions{})
+	return &model, diags
+}
+
+func resolveReadResponse(ctx context.Context, client *clients.ResourceClient, azureResourceId, azureResourceType, apiVersion string, requestBody map[string]interface{}, getResponse interface{}, readOverride *ReadOverrideModel, requestOptions clients.RequestOptions) (interface{}, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	if readOverride == nil {
+		return autoReadResponse(ctx, client, azureResourceId, azureResourceType, apiVersion, requestBody, getResponse, requestOptions)
+	}
```
gerrytan: If there is no explicit override provided in the HCL config, there should not be any automated fallback. This is a surprising behaviour and users without the correct permission can unexpectedly face a 403 error.

gerrytan (dismissed):
@jaredfholgate I've redesigned the implementation into a schema instead of 3 separate prop, and also merged latest main.

I've got 1 more design feedback below. Please let me know if you agree, and I can push the modification.

internal/services/read_action.go
jaredfholgate: This was by design. Expecting an inexperienced user to know they need to provide this setting is not user friendly. The whole PR was about attempting to fix the underlying problem without expecting the user to figure it out themselves. The input is there for advanced scenarios where we can't figure out the API to call.

If we are missing some context in a permissions error message, then we should add that of course.

Setting the read action to `none` is what reverts to existing behaviour.

jaredfholgate: I guess now it is an object the read action being set to `none` is a less obvious way to set it to existing behaviour though. Perhaps we need something more obvious there?

jaredfholgate: Perhaps an enabled property that defaults to true instead?

```hcl
read_override = { 
  enabled       = true // defaults to `true`. If the other attributes are not set we use a default  `POST /list` as a fallback after attempting a `GET`
  method        = "POST" // required, POST is the only valid value
  action        = "list" // required
  response_path = "..." // optional
}
```

gerrytan:
@jaredfholgate as discussed internally in Teams, I've modified this PR as follow:

1. Only update_resource has read_override, will add it to azapi_resource when the use case arise
2. No implicit fallback, user has to explicitly set the read_override block in the config. AzAPI is designed to be a thin provider on top of REST API, it shouldn't contain heavy implicit logic such as switching the HTTP method used. Also user unaware could get caught with permission issue.
3. Dropped the path property, I don't see this being used in the use case provided. Will consider to re-add this if needed.

I'll also raise a follow up PR to update the examples of managing Microsoft.Web/sites/config properties using update_resource.

acctest: https://dev.azure.com/azclitools/internal/_build/results?buildId=346542&view=results

internal/services/azapi_update_resource.go
```diff
@@ -599,7 +602,12 @@ func (r *AzapiUpdateResource) Read(ctx context.Context, request resource.ReadReq
 		QueryParameters: clients.NewQueryParameters(common.AsMapOfLists(model.ReadQueryParameters)),
 	}
 	requestOptions.RetryOptions, requestOptions.LastRetryError = clients.NewRetryOptions(model.Retry)
-	responseBody, err := client.Get(ctx, id.AzureResourceId, id.ApiVersion, requestOptions)
+	readOverride, readDiags := readOverrideFromObject(ctx, model.ReadOverride)
+	response.Diagnostics.Append(readDiags...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+	responseBody, err := readResource(ctx, client, id.AzureResourceId, id.ApiVersion, readOverride, requestOptions)
```
hqhqhqhqhqhqhqhqhqhqhq, line 618: Could we avoid automatically saving secrets from POST/list into `output`?

With default provider settings and no `response_export_values`, the appsettings dictionary is copied into output. That includes values supplied through the write-only `sensitive_body`:

```text
Input:         sensitive_body.properties.API_KEY = <secret>
After refresh: output.properties.API_KEY = <secret>
```

The secret ends up back in Terraform state, and output is not marked sensitive. Could we prevent this automatic export and add a test for this case? Marking output sensitive alone would still save the value in state.

internal/services/azapi_update_resource.go
```diff
@@ -599,7 +602,12 @@ func (r *AzapiUpdateResource) Read(ctx context.Context, request resource.ReadReq
 		QueryParameters: clients.NewQueryParameters(common.AsMapOfLists(model.ReadQueryParameters)),
 	}
 	requestOptions.RetryOptions, requestOptions.LastRetryError = clients.NewRetryOptions(model.Retry)
-	responseBody, err := client.Get(ctx, id.AzureResourceId, id.ApiVersion, requestOptions)
+	readOverride, readDiags := readOverrideFromObject(ctx, model.ReadOverride)
```
hqhqhqhqhqhqhqhqhqhqhq, line 613: Could we use `read_override` for the reads before and after PUT too? It currently only changes refresh; apply still uses GET.

For appsettings, GET returns empty properties and PUT replaces all settings. So even with the override configured:

```text
Azure has:         KEEP=existing, CHANGE=old
Terraform manages: CHANGE=new
After apply:       CHANGE=new  (KEEP is deleted)
```

This leaves the existing deletion problem unfixed. The GET after PUT also leaves exported settings empty until refresh. Could we use the override for both reads and test that an existing setting outside our config survives?

internal/services/azapi_update_resource.go
gerrytan: The sensitive body is not saved to state / exported to output. I've verified using an acctest added in 26da492d9 and it passes locally.

internal/services/azapi_update_resource.go
gerrytan: Good catch @hqhqhqhqhqhqhqhqhqhqhq , fixed in 39b97e9d8 and added test coverage. Still waiting for the latest acctest to get back. I'll send another review request when ready.

https://dev.azure.com/azclitools/internal/_build/results?buildId=346900&view=results

internal/services/azapi_update_resource.go
gerrytan: Wait, after fixing the read logic, I can see the problem now: read_override is clashing with sensitive_body. The problem here is we have [an API that uses POST /list for "read" but it's not designed / documented explicitly for secret / sensitive value](https://learn.microsoft.com/en-us/rest/api/appservice/web-apps/list-application-settings?view=rest-appservice-2026-07-15&tabs=HTTP), so users are using it for regular values too.

I have refactored read_override to be incompatible with sensitive_body in ecda6dd35.

Will ping you when the acctest result come back. Good catch @hqhqhqhqhqhqhqhqhqhqhq , thx!

https://dev.azure.com/azclitools/internal/_build/results?buildId=346972&view=results

gerrytan:
@hqhqhqhqhqhqhqhqhqhqhq all feedbacks have been addressed and acctest result is good https://dev.azure.com/azclitools/internal/_build/results?buildId=346972&view=results. Can you please review this PR again. Thanks!

hqhqhqhqhqhqhqhqhqhqhq (approved):

PR #1174, opened 2026-07-17
`azapi_data_plane_resource` - support key vault certificate

LingyuTang:
### Description

This PR is to support key vault certificate in `azapi_data_plane_resource`. Different from other data plane resource, certificate using PATCH for update and needs to request for another endpoint for creation status, so it needs customization.

[API doc](https://learn.microsoft.com/en-us/rest/api/keyvault/certificates/create-certificate/create-certificate?view=rest-keyvault-certificates-2025-07-01&tabs=HTTP)
[Acc test](https://dev.azure.com/azclitools/internal/_build/results?buildId=330439&view=results)


gerrytan (approved):
Nice work @LingyuTang , thanks for linking the API doc and running acctest. This PR looks good to me.

PR #1053, opened 2026-02-19
feat: add AI Foundry assistants support with customization and resour…

roder:
## Description

This PR adds support for **Azure AI Foundry Agents** (assistants, threads, messages, runs, vector stores, etc.) to the `azapi_data_plane_resource` via a new customization pattern that handles server-generated resource identifiers.

### Problem

AI Foundry Agents APIs follow a **POST-to-collection** pattern where the service generates the resource identifier (e.g., `asst_abc123` for assistants) instead of accepting a client-specified name. This is incompatible with the existing `azapi_data_plane_resource` contract, which requires users to specify a `name` attribute that becomes part of the resource ID before creation.

### Solution

This PR introduces an **optional extension interface** (`DataPlaneResourceWithCreateResult`) that allows customizations to:
1. POST to a collection endpoint (e.g., `/assistants`)
2. Extract the server-generated `id` from the response
3. Return the actual resource ID for subsequent read/update/delete operations

This follows the established **KeyVault customization pattern** but extends it to support APIs with server-generated identifiers.

---

### 1. Extended Customization Interface

**File:** `internal/services/customization/interface.go`

- Added optional `CreateResultFunc()` method that returns `(DataPlaneResourceId, responseBody, error)`
- Maintains backward compatibility — existing customizations (KeyVault) are unaffected
- Allows customizations to override the resource ID construction after create

### 2. Updated `azapi_data_plane_resource` Schema & Logic

**File:** `internal/services/azapi_data_plane_resource.go`

**Schema changes:**
- `name` attribute is now **optional + computed** (previously required)
- Added `UseStateForUnknown()` plan modifier so Terraform knows to expect a server value
- Updated description to clarify server-generated ID behavior

**Create/update flow changes:**
- For resources with `CreateResultFunc()`, allows omitting `name` during plan
- Uses a placeholder (`__generated__`) to construct the initial request path
- After creation, swaps the placeholder with the actual server-returned ID
- Skips the "pre-GET exists check" for server-generated ID resources (the ID doesn't exist yet)
- Enforces `name` required for all other data plane resource types via runtime validation

### 3. AI Foundry Assistant Customization

**File:** `internal/services/customization/ai_foundry_assistant_customization.go`

Implements the full CRUD lifecycle for `Microsoft.AIFoundry/agents/assistants@v1`:

| Operation | Method | Endpoint | Notes |
|-----------|--------|----------|-------|
| **Create** | `POST` | `{parent}/assistants` | Returns `{"id": "asst_..."}` in response |
[...]

PR #1178, opened 2026-07-21
`azapi_data_plane_resource` - allow `name` to be optional based on URL format

LingyuTang:
## Description

This PR:

1. Uses the data plane API format configuration to determine whether the name property is required for a given resource.
2. Performs a small refactor by moving the logic that determines whether the name property is returned by the service into the getCreateResultFunc function.

### Background 
For singleton resource like key vault contact, the data plane api doesn't require `name` property ([api docuementation](https://learn.microsoft.com/en-us/rest/api/keyvault/certificates/set-certificate-contacts/set-certificate-contacts?view=rest-keyvault-certificates-2025-07-01&tabs=HTTP)). However, during the [PR #1053](https://github.com/Azure/terraform-provider-azapi/pull/1053/changes), it introduces the guard toward `name` property that will reject any empty/unset name, which cause the [issue #1091](https://github.com/Azure/terraform-provider-azapi/issues/1091)

Acc test https://dev.azure.com/azclitools/internal/_build/results?buildId=331460&view=results. Failed tests also failed on main and doesn't seem to be related to this PR.

### DIFF
--- a/internal/services/azapi_data_plane_resource.go
+++ b/internal/services/azapi_data_plane_resource.go
@@ -432,10 +432,6 @@ func (r *DataPlaneResource) CreateUpdate(ctx context.Context, requestConfig tfsd
 	if isNewResource && hasCreateResult {
 		resourceName = "__generated__"
 	}
-	if resourceName == "" {
-		diagnostics.AddError("Invalid configuration", `The argument "name" must be set for this resource type.`)
-		return
-	}
 
 	id, err := parse.NewDataPlaneResourceId(resourceName, plan.ParentID.ValueString(), plan.Type.ValueString())
 	if err != nil {
@@ -613,6 +609,10 @@ func getCreateResultFunc(config *DataPlaneResourceModel) (customization.CreateRe
 
 func validateDataPlaneResourceName(config *DataPlaneResourceModel) error {
 	if config == nil || config.Type.IsNull() || config.Type.IsUnknown() {
+		return nil
+	}
+
+	if !parse.HasNameSegment(config.Type.ValueString()) {
 		return nil
 	}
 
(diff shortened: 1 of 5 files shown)


internal/services/azapi_data_plane_resource.go
```diff
@@ -607,30 +593,42 @@ func (r *DataPlaneResource) CreateUpdate(ctx context.Context, requestConfig tfsd
...
 	}
 
-	customizedResource := customization.GetCustomization(config.Type.ValueString())
-	hasCreateResult := false
-	if customizedResource != nil {
-		if v, ok := (*customizedResource).(customization.DataPlaneResourceWithCreateResult); ok && v.CreateResultFunc() != nil {
-			hasCreateResult = true
-		}
+	if !parse.HasNameSegment(config.Type.ValueString()) {
+		return nil
```
gerrytan: There's a bug here, if there is no name segment, but the TF config supplies `name = "foo"`, this validation logic return too early and it will get stored into state, and cause subsequent plan diff.

Can you please add an acctest coverage for this. For such data plane type, we should expect a plan-time validation error if user supplied a name value.

gerrytan (changes requested):

internal/services/azapi_data_plane_resource.go
LingyuTang: Hey Gerry, thanks for the review. I have updated the logic so that when user enter name for the data plane resource when it is not required it will error out instead of early return and the add the acc test for it. I rerun the pipeline as well, the new test pass and failed ones are policy related. See the details in pr description.

gerrytan (approved):
Thank you for addressing the feedback, adding test coverage, and running the test again. This PR now looks good to me @LingyuTang

PR #1180, opened 2026-07-22
fix: temporarily suppress invoke policy acctest failure

gerrytan:
This PR temporarily suppress the failure on invoke policy acctest. This is needed because the feature cannot be enabled on the new acctest tenant at least for another 2 weeks.

## Testing

Check `TestAccResourceWithAcquirePolicyToken_` acctest passes: https://dev.azure.com/azclitools/internal/_build/results?buildId=331597&view=results

### DIFF
--- a/internal/services/azapi_resource_acquire_policy_token_test.go
+++ b/internal/services/azapi_resource_acquire_policy_token_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"regexp"
 	"testing"
+	"time"
 
 	"github.com/Azure/terraform-provider-azapi/internal/acceptance"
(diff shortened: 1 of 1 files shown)


LingyuTang (approved):
LGTM! We probably need to set up a notification for cleaning up this later 🤔

gerrytan:
Thanks! Acctest result look good: https://dev.azure.com/azclitools/internal/_build/results?buildId=331597&view=logs&j=ca395085-040a-526b-2ce8-bdc85f692774&t=632617cc-49d1-5f82-37d2-600ea5832b67, the single failure `--- FAIL: TestAccGenericResource_preflightChildResource ` is not related and I'll address it in a separate PR

Issue #1138, opened 2026-06-03
feat: support write-only request headers (excluded from state and the terraform plan)


jaredfholgate:
# Support write-only request headers (excluded from state and the Terraform plan)

## Use case (generic module development)

Module authors often want to attach lightweight usage telemetry to the Azure requests their module makes — for example a header carrying the module name and version — so module usage can be measured from standard **ARM request logs** without standing up a separate telemetry provider or pipeline. Today custom request headers can be set on `azapi_resource`, but their values are **rendered in the Terraform plan**. For generic module development this causes:

- **Plan noise:** the header appears as a tracked attribute and produces a diff every time its value changes (e.g. on every module version bump), even when no real infrastructure change occurred.
- **Poor consumer experience:** plans show churn unrelated to the actual resource.
- **Pressure to avoid telemetry**, or to fall back to heavier mechanisms — a dedicated telemetry provider that deploys extra resources/data sources, bloating state and requiring consumers to trust an additional provider.

We need a way to set request headers that are **write-only**: sent on the API request, but never persisted to state nor shown in the plan diff.

## High-level implementation options

1. **Write-only header argument.** Add a `write_only_headers` argument (or a write-only flag on existing headers) on `azapi_resource`/`azapi_update_resource` that uses Terraform's write-only / ephemeral value support, so values are sent on apply but never stored in state or surfaced in the plan.
2. **Non-persisted telemetry header.** Add a dedicated usage-header argument the provider treats as non-persisted metadata — explicitly excluded from diff computation — specifically for attaching module-identity headers to ARM requests.

## Example configuration

```hcl
# Inside a reusable module
resource "azapi_resource" "example" {
  type      = "Microsoft.Storage/storageAccounts@2023-01-01"
  name      = var.name
  parent_id = var.resource_group_id

  body = {
    properties = {
      minimumTlsVersion = "TLS1_2"
    }
  }

  # Proposed: write-only headers — sent on the API request, but never
  # persisted to state and never shown in the plan diff. Today these same
  # values would render in the plan and churn on every version bump.
  write_only_headers = {
    "X-Module-Id"      = "storage-account"
    "X-Module-Version" = "0.5.1"
  }
}
```

## Optional advanced scenario — automatic module-metadata telemetry

[...]

PR #1182, opened 2026-07-24
feat: add telemetry headers

gerrytan:
Implements #1138 where a new optional block is added to all resources and data sources for AVM modules to be able to report analytics.:

```
telemetry_headers {
  module_id      = "some-module-name"
  module_version = "0.0.1"
}
```

The headers `X-MODULE-ID` and `X-MODULE-VERSION` will then be sent in the Create and Update only http requests. Due to the design of HC framework where we can't see the config during Read and Delete, so it's not supported there.

This is write-only and doesn't get stored into state. Previously analytics was done using `create_headers`, `update_headers` etc. but it created unecessary plan diff when module version is bumped.

It is deliberately designed to allow only specific headers to reduce the security risk.

## Testing

- Added unit test and acctest coverage
- Locally ran acctest with mitmdump to inspect the header do get sent
- Pipeline acctest: https://github.com/Azure/terraform-provider-azapi/pull/1182/checks?check_run_id=89375189951

### DIFF
--- a/docs/actions/resource_action.md
+++ b/docs/actions/resource_action.md
@@ -140,7 +140,7 @@ action "azapi_resource_action" "power_on" {
 - `body` (Dynamic) A dynamic attribute that contains the request body.
 
 	-> Ensure the dynamic value is not a string.
-- `headers` (Map of String) Headers to include in the request.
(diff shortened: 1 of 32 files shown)


gerrytan:
Closing as the preferred solution is to use dedicated telemetry provider to collect analytics: https://github.com/Azure/terraform-provider-modtm

PR #1183, opened 2026-07-27
Preserve azapi_resource state for in-flight Create interrupted by SIGINT

### LINKED ISSUE #1110: azapi_resource: in-flight Create not preserved in state when terraform receives SIGINT (resource left orphaned in Azure)
### Summary

When `terraform apply` is interrupted with SIGINT (Ctrl+C, CI cancellation, job timeout, ...) while an `azapi_resource` is mid-Create, the resource ends up **created in Azure but absent from terraform state**. The next apply then fails with `A resource with the ID "<id>" already exists - to be managed via Terraform this resource needs to be imported`, requiring a manual `terraform import` to recover.

This is the azapi counterpart of [hashicorp/terraform-provider-azurerm#32230](https://github.com/hashicorp/terraform-provider-azurerm/issues/32230) - I confirmed the same pattern in azapi v2.9.0.

### Versions

- Terraform: `v1.14.9` on `darwin_arm64`
- Provider: `Azure/azapi v2.9.0`

### Repro

A small azapi-only config that creates a resource with a 20-30s ARM Create:

```hcl
terraform {
  required_version = "~> 1.7"
  required_providers {
    azapi = { source = "Azure/azapi", version = "~> 2.0" }
  }
}

provider "azapi" {
  subscription_id = "00000000-0000-0000-0000-000000000000"
}

resource "azapi_resource" "sa" {
  type      = "Microsoft.Storage/storageAccounts@2023-01-01"
  name      = "stcanceltestsample"
  location  = "westeurope"
  parent_id = "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/example-rg"

  body = {
    sku  = { name = "Standard_LRS" }
    kind = "StorageV2"
    properties = {
      accessTier               = "Hot"
      allowBlobPublicAccess    = false
      minimumTlsVersion        = "TLS1_2"
      supportsHttpsTrafficOnly = true
    }
  }
}
```

Steps:

1. `terraform apply -auto-approve`.
2. Once `azapi_resource.sa: Creating...` appears, wait ~4 seconds (so the ARM PUT has been accepted and the provider is in the poll loop).
3. Send SIGINT to the terraform process (`kill -INT <pid>` or Ctrl+C).

### Actual behaviour

Terraform exits non-zero with:

```
azapi_resource.sa: Creating...
azapi_resource.sa: Still creating... [00m10s elapsed]
Stopping operation...

Interrupt received.
Please wait for Terraform to exit or data loss may occur.
Gracefully shutting down...

Error: execution halted

Error: Failed to create/update resource

  with azapi_resource.sa,
  on main.tf line 25, in resource "azapi_resource" "sa":
  25: resource "azapi_resource" "sa" {

creating/updating Resource: (ResourceId
".../providers/Microsoft.Storage/storageAccounts/stcanceltestsample"
/ Api Version "2023-01-01"): context canceled
```

After exit:

| Check                         | Result                                          |
| ----------------------------- | ----------------------------------------------- |
| `terraform state list`        | empty                                           |
| `az storage account show ...` | `provisioningState: Succeeded`, full id present |

ARM accepted the PUT and finished the create server-side, but the provider cancelled its poll context on the terraform-core Stop signal and never returned a resource id, so core had nothing to write into state. The resource is orphaned.

### Expected behaviour

On first SIGINT, terraform core asks the provider to **gracefully stop** while it flushes state (see `apparentlymart`'s walkthrough on [hashicorp/terraform#34266](https://github.com/hashicorp/terraform/issues/34266)). For an in-flight LRO Create, "graceful" should mean either:

1. Wait for the in-flight ARM Create poll to complete and return the resource id, so core can persist it (preferred for short-to-medium-running creates such as storage accounts), or
2. Persist the resource id as soon as ARM has accepted the PUT (i.e. before the poll loop completes), so even an immediate hard abort leaves a resource that can be reconciled on the next apply rather than orphaned.

Either path is far better than the current behaviour, which leaves the user with a resource in Azure but no state record and no graceful recovery short of `terraform import`.

### Notes / related

- Same pattern in azurerm: [hashicorp/terraform-provider-azurerm#32230](https://github.com/hashicorp/terraform-provider-azurerm/issues/32230). Fixes on both providers should ideally share the same approach since they face the same upstream contract.
- Same `context done` codepath family as [#1078](https://github.com/Azure/terraform-provider-azapi/issues/1078) (the **deadline** case, not the **cancel** case). A single fix that preserves state for in-flight LROs across both `context.DeadlineExceeded` and `context.Canceled` would cover both issues.
- Mid-Update and mid-Delete cancels are explicitly out of scope for this issue (matching the scope of azurerm#32230).

### Impact

This particularly affects CI/CD workflows where builds are frequently cancelled to iterate quickly, since every cancelled build that hits an in-flight Create produces an orphan that has to be cleaned up manually.

gerrytan:
Fixes #1110. Also covers the `context.DeadlineExceeded` case from #1078.

## Problem

When `terraform apply` is interrupted with SIGINT (Ctrl+C, CI cancellation, job timeout) while an `azapi_resource` Create is mid-flight, the resource ends up **created in Azure but absent from state**. The next apply then fails with `A resource with the ID "<id>" already exists ... needs to be imported`, requiring a manual `terraform import` to recover.

## Root cause

In `AzapiResource.CreateUpdate` (`internal/services/azapi_resource.go`), when `CreateOrUpdate` fails it already attempts a recovery `Get` to still capture the resource id for a new resource. But that recovery `Get` — and the `responseState.Set` that follows — reused the **same context** that SIGINT had already cancelled (`context.Canceled`), or that had expired (`context.DeadlineExceeded`, #1078). So the recovery read failed too, no id was captured, and terraform wrote nothing to state.

On first SIGINT, terraform core cancels the RPC context but still flushes whatever state the provider returns — so if the provider returns the id alongside the error, core persists it and the next apply reconciles instead of erroring.

## Fix

Run the recovery `Get` (and the subsequent state write) on a context **detached** from the cancelled/expired parent when the create fails with `context.Canceled` or `context.DeadlineExceeded`:

```go
recoveryCtx := ctx
if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
    var cancel context.CancelFunc
    recoveryCtx, cancel = context.WithTimeout(context.WithoutCancel(ctx), 5*time.Minute)
    defer cancel()
}
```

- `context.WithoutCancel` drops cancellation but preserves values (auth, logging fields); the bounded timeout prevents a hung read from blocking terraform shutdown.
- Non-context errors keep the original behaviour unchanged.
- If the resource genuinely was not created (cancel before ARM accepted the PUT), the recovery read returns 404, so no false state is written.

## Testing

Adds `internal/clients/resource_client_sigint_test.go`, a no-Azure test using a fake ARM transport that:
1. Accepts the PUT (resource created server-side) and returns an LRO poller.
2. Cancels the context mid-poll (simulating SIGINT) so `CreateOrUpdate` returns `context.Canceled`.
3. Asserts a recovery `Get` on the **same cancelled** context fails (the orphaning bug).
4. Asserts a recovery `Get` on a **detached** context succeeds and returns the id (the fix).

Also ran acctest: https://github.com/Azure/terraform-provider-azapi/pull/1183/checks?check_run_id=89890752648

### DIFF
--- a/internal/clients/resource_client_sigint_test.go
+++ b/internal/clients/resource_client_sigint_test.go
@@ -1,0 +1,196 @@
+package clients
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+)
+
+// sigintTransport simulates ARM while an azapi_resource Create is interrupted by
+// SIGINT, reproducing https://github.com/Azure/terraform-provider-azapi/issues/1110.
+//
+// Sequence modelled:
+//  1. PUT (CreateOrUpdate) is accepted by ARM: the resource is created
+//     server-side and a long-running-operation (LRO) poller is returned via the
+//     Azure-AsyncOperation header.
+//  2. While the provider polls the LRO, terraform core receives SIGINT and
+//     cancels the context (simulated here by cancelling on the first poll GET).
+//     The poll therefore fails with context.Canceled.
+//
+// Like the real net/http transport, Do returns ctx.Err() when the request's
+// context is already cancelled. This is what makes the provider's post-failure
+// recovery Get in azapi_resource.go (which reuses the same cancelled context)
+// fail as well, leaving the resource orphaned in Azure but absent from state.
+type sigintTransport struct {
+	host         string
+	resourcePath string
+	asyncOpPath  string
+	cancel       context.CancelFunc
+
+	mu        sync.Mutex
+	created   bool
+	pollCount int
+}
+
+func (t *sigintTransport) Do(req *http.Request) (*http.Response, error) {
+	if err := req.Context().Err(); err != nil {
+		return nil, err
+	}
+
+	switch {
+	case req.Method == http.MethodPut && req.URL.Path == t.resourcePath:
+		// ARM accepts the create and starts an async operation.
+		t.mu.Lock()
+		t.created = true
+		t.mu.Unlock()
+		resp := &http.Response{
+			StatusCode: http.StatusCreated,
+			Header:     http.Header{"Content-Type": []string{"application/json"}},
+			Body: io.NopCloser(strings.NewReader(
+				`{"id":"` + t.resourcePath + `","properties":{"provisioningState":"Creating"}}`)),
+			Request: req,
+		}
+		resp.Header.Set("Azure-AsyncOperation", t.host+t.asyncOpPath)
+		return resp, nil
+
+	case req.Method == http.MethodGet && req.URL.Path == t.asyncOpPath:
+		// SIGINT arrives mid-poll: cancel the context exactly as terraform core
+		// would when the operation is stopped, then report "still in progress".
+		// azcore observes the cancellation and PollUntilDone returns context.Canceled.
+		t.mu.Lock()
+		t.pollCount++
+		t.mu.Unlock()
+		if t.cancel != nil {
+			t.cancel()
+		}
+		return &http.Response{
+			StatusCode: http.StatusOK,
+			Header:     http.Header{"Content-Type": []string{"application/json"}},
+			Body:       io.NopCloser(strings.NewReader(`{"status":"InProgress"}`)),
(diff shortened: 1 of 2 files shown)


LingyuTang (approved):
Looks cool to me. Thanks Gerry🙏

PR #1189, opened 2026-08-03
VCR testing for acctests

gerrytan:
This PR introduces VCR testing that will gradually replace majority of acctests. We still want to keep a small set of acctests to ensure upstream problem are discovered early.

See the developer info added in the README.md

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -108,6 +108,7 @@ testacc: fmtcheck
 acctests: fmtcheck
 	TF_ACC=1 go test -v ./internal/services $(TESTARGS) -timeout $(TESTTIMEOUT) -ldflags="-X=github.com/Azure/terraform-provider-azapi/version.ProviderVersion=acc"
 
+# Go accepts multiple -run flags and uses the last one, allowing CI to run all VCR tests while TESTARGS can select an individual test locally.
 vcrreplay: fmtcheck
 	AZAPI_VCR_MODE=replay TF_ACC=1 go test -v ./internal/services -run 'TestVcr' $(TESTARGS) -timeout $(TESTTIMEOUT) -ldflags="-X=github.com/Azure/terraform-provider-azapi/version.ProviderVersion=acc"
 
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -108,6 +108,12 @@ testacc: fmtcheck
 acctests: fmtcheck
 	TF_ACC=1 go test -v ./internal/services $(TESTARGS) -timeout $(TESTTIMEOUT) -ldflags="-X=github.com/Azure/terraform-provider-azapi/version.ProviderVersion=acc"
 
+vcrreplay: fmtcheck
+	AZAPI_VCR_MODE=replay TF_ACC=1 go test -v ./internal/services -run 'TestVcr' $(TESTARGS) -timeout $(TESTTIMEOUT) -ldflags="-X=github.com/Azure/terraform-provider-azapi/version.ProviderVersion=acc"
+
+vcrrecord: fmtcheck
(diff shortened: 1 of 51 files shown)


GNUmakefile
```diff
@@ -108,6 +108,12 @@ testacc: fmtcheck
 acctests: fmtcheck
 	TF_ACC=1 go test -v ./internal/services $(TESTARGS) -timeout $(TESTTIMEOUT) -ldflags="-X=github.com/Azure/terraform-provider-azapi/version.ProviderVersion=acc"
 
+vcrreplay: fmtcheck
+	AZAPI_VCR_MODE=replay TF_ACC=1 go test -v ./internal/services -run 'TestVcr' $(TESTARGS) -timeout $(TESTTIMEOUT) -ldflags="-X=github.com/Azure/terraform-provider-azapi/version.ProviderVersion=acc"
```
magodo, line 113: With `-run 'TestVcr'`, we can't manually select which test case to record by running this make target?

README.md
```diff
@@ -180,6 +194,42 @@ The `ARM_READER_CLIENT_ID` and `ARM_READER_CLIENT_SECRET` follow the same proces
...
+
+#### Recording a cassette
+
+Recording runs live, so it needs the same credentials and Azure quota as an acceptance test (see the environment variables above):
+
+```sh
+make vcrrecord TESTARGS='-run TestVcrAccGenericResource_basic'
+```
+
+This writes `internal/services/testdata/cassettes/TestVcrAccGenericResource_basic.yaml`. Review it (it must contain only canonical placeholder IDs and no secrets / sensitive data) and commit it alongside the test.
```
magodo: The review part is challenging given the big size of the sequence. Maybe we can provide a skill to let AI to do that review instead?

internal/services/testdata/cassettes/TestVcrAccGenericResource_basic.yaml
```diff
@@ -0,0 +1,1039 @@
+---
```
magodo, line 1: In terms of size and security concerns below:

- Size: The size of the record files can easily blow up the repo
- Security: This file can easily leak security information to public once the commit/PR is submit to the repo (even if it is later reset/overwritten)

Also since the purpose of the replay-based acctest is mainly for our own verification purpose (e.g. enforcing in CI of this repo). We might consider to put it in a private repo, where we only clone and use it when recording and when running in the CI for this repo?

GNUmakefile
gerrytan: We can pass multiple `-run` arg and only the last one wins. The intent is `make vcrreplay` will run all `TestVcr*` tests (I need to fix the regex to be `^TestVcr`), but you can run `TESTARGS='-run="TestVcrXyz"' make vcrreplay` if you just want to run a single one.

I'll add an explanation comment in code and fix the regex pattern.

internal/services/testdata/cassettes/TestVcrAccGenericResource_basic.yaml
gerrytan: Agree on the security concern of accidentally pushing sensitive data (name, subId, etc.), but I prefer the casette / recording to be public so any contributor can run and reproduce issue locally.

Have a look at `func sanitizeInteraction`, sensitive data should be sanitized here. But it is indeed a concern if new kind of prop / header not yet sanitized appear and we forgot to update this.

I'll investigate implementing some static check to prevent such sensitive data from being pushed.

README.md
gerrytan: The intent is just to ensure no sensitive data accidentally pushed, I'll rephrase this.

I don't think AI skill is necessary as vanilla "check if this file contains any sensitive information" prompt works well even with free-tier models.

gerrytan: 49d0df895

GNUmakefile
gerrytan: 3376c010

PR #1190, opened 2026-08-03
Strip UTF-8 BOM prefix from JSON responses

### LINKED ISSUE #1098: `azapi_resource` fails to read resources when Azure API returns UTF-8 BOM in JSON response
## Description

We're managing Azure Peering resources (`Microsoft.Peering/peerings`) via `azapi_resource` and since yesterday all plan/apply operations fail with:

```sh
Error: reading Resource: (ResourceId "/.../Microsoft.Peering/peerings/{name}"
/ Api Version "2025-05-01"): unmarshalling type *interface {}: invalid character 'ï'
looking for beginning of value
```

The `ï` is `0xEF` - first byte of a UTF-8 BOM (`EF BB BF`). The Azure Peering API is returning a 3 byte BOM prefix before the JSON body. Go's `encoding/json` chokes on it.

The Azure Go SDK (`azcore/runtime/response.go`) already handles this - there's a `removeBOM` function that does `bytes.TrimPrefix(b, []byte("\xef\xbb\xbf"))` before unmarshalling. But `azapi_resource` apparently doesn't go through that path. So `azurerm` would survive this just fine, but `azapi` doesn't.

## Reproduction

```bash
az rest --method GET \
  --url "https://management.azure.com/subscriptions/{sub}/resourceGroups/{rg}/providers/Microsoft.Peering/peerings/{name}?api-version=2025-05-01" \
  --output-file /tmp/peering.bin

xxd -l 4 /tmp/peering.bin
# Output: 00000000: efbb bf7b    ...{
# First 3 bytes are UTF-8 BOM, then `{` starts the JSON
```

Then any `terraform plan` on an `azapi_resource` of type `Microsoft.Peering/peerings@2025-05-01` fails during state refresh.

## Suggested fix

Strip BOM before JSON decode, same as the Go SDK already does:

```go
body = bytes.TrimPrefix(body, []byte("\xef\xbb\xbf"))
```

This would make the provider resilient to any Azure API returning BOM - not just Peering. It's a known Azure-side pattern (see Azure/msrest-for-python#144, Azure/azure-sdk-for-java#5475).

## Environment

- `azapi` provider: tested v2.5.0, v2.8.0, v2.9.0 — all fail identically
- Terraform: 1.14.3
- Terragrunt: 0.99.5
- OS: Ubuntu (GitHub Actions runner) + macOS (local)

gerrytan:
Fixes #1098 by introducing a http transport policy that strips UTF-8 BOM prefix on JSON response payload. Go net/http deliberately does not strip the UTF-8 BOM because they prefer downstream packages to do it.

## Testing

1. I added an acctest that rely on mitmproxy since actual Microsoft.Peering is difficult to setup, this test only works locally because I'm not sure how reliable it will be in ADO pipeline
2. Run acctest to ensure no regression: https://github.com/Azure/terraform-provider-azapi/pull/1190/checks?check_run_id=91622596732

### DIFF
--- a/internal/clients/client.go
+++ b/internal/clients/client.go
@@ -65,6 +65,7 @@ func (client *Client) Build(ctx context.Context, o *Option) error {
 	}
 	perRetryPolicies := make([]policy.Policy, 0)
 	perRetryPolicies = append(perRetryPolicies, NewLiveTrafficLogPolicy())
+	perRetryPolicies = append(perRetryPolicies, newUTF8BOMPolicy())
 
 	allowedHeaders := []string{
 		"Access-Control-Allow-Methods",
--- a/internal/clients/utf8_bom_policy.go
+++ b/internal/clients/utf8_bom_policy.go
@@ -1,0 +1,54 @@
+package clients
+
+import (
+	"bufio"
+	"bytes"
+	"io"
+	"mime"
+	"net/http"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+)
+
+var utf8BOM = []byte{0xef, 0xbb, 0xbf}
+
+type utf8BOMPolicy struct{}
+
+func newUTF8BOMPolicy() policy.Policy {
+	return &utf8BOMPolicy{}
+}
+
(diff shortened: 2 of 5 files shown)


LingyuTang (approved):
Seems good to me! Thanks Gerry

PR #1192, opened 2026-08-05
azapi_resource: re-introduce ignore_body_changes property

### LINKED ISSUE #1137: feat: support dynamic, value-driven `ignore_changes` in `azapi_resource`
# Support dynamic, value-driven `ignore_changes` in `azapi_resource`

## Use case (generic module development)

When building reusable Terraform modules on top of `azapi_resource`, module authors cannot expose ignore behaviour as a configurable input. Terraform's `lifecycle.ignore_changes` meta-argument only accepts a **static list of literal attribute paths** — it cannot reference variables, locals, or computed values. For generic module development this means:

- A module cannot let its consumers opt in or out of ignoring specific body fields (for example properties mutated out-of-band by Azure Policy, autoscalers, or other controllers).
- Authors must hard-code ignore paths, forcing one-size-fits-all behaviour that produces spurious diffs for some consumers and drift for others.
- The classic case is **tags**: some consumers want them managed externally and ignored, others want them enforced by the module. With a static `ignore_changes` there is no clean way to make this a module input.

AzAPI previously offered `ignore_body_changes`, which gave dynamic, value-driven control over which parts of the request body were ignored. Its deprecation (#583) removed the only mechanism that let module authors **parameterise** this behaviour, so a provider-native dynamic equivalent is needed.

## High-level implementation options

1. **Reinstate a dynamic body-ignore argument.** Provide a regular provider argument (e.g. `ignore_changes` / `ignore_body_changes`) that accepts a list of JSON body paths sourced from a variable. Because it is a normal argument rather than a meta-argument, it can reference variables/locals and can therefore be driven by module inputs.
2. **Path-expression based ignore.** Accept JMESPath/path expressions over the resource body, evaluated by the provider at plan time, again as a dynamic-capable argument so module authors can compose the ignore set from configuration.

## Example configuration

```hcl
# Inside a reusable module
variable "ignore_changes" {
  type        = list(string)
  default     = []
  description = "JSON body paths the module consumer wants Terraform to ignore (e.g. [\"tags\", \"properties.encryption\"])."
}

  type      = "Microsoft.Storage/storageAccounts@2023-01-01"
  name      = var.name
  parent_id = var.resource_group_id

  body = {
    properties = {
      minimumTlsVersion = "TLS1_2"
    }
    tags = var.tags
  }

  # Proposed: a dynamic, variable-driven ignore list.
  # This is NOT possible today with lifecycle.ignore_changes, which only
  # accepts a static list of literal paths and rejects variables/locals.
  ignore_changes = var.ignore_changes
}
```

```hcl
# Consumer A — let an external process manage tags
module "storage" {
  source         = "./modules/storage"
  name           = "examplesa"
  ignore_changes = ["tags"]
}

# Consumer B — enforce everything (default behaviour, nothing ignored)
module "storage" {
  source = "./modules/storage"
  name   = "examplesa"
}
```

## Note on the underlying gap

The root cause is a gap in **Terraform core** — the `lifecycle.ignore_changes` meta-argument cannot be dynamic (it rejects variables and computed values). This has been raised upstream repeatedly with **no traction from IBM (HashiCorp)** toward a fix. A provider-side solution in AzAPI is therefore the pragmatic path forward.

*Related:* Azure/terraform-provider-azapi#583 (Deprecation of `ignore_body_changes`).

gerrytan:
Introduces `ignore_body_changes` to `azapi_resource` for diff suppression.

Example usage:

```hcl
variable "ignore_tags" {
  type = bool
}

resource "azapi_resource" "test" {
  type      = "Microsoft.Network/virtualNetworks@2024-05-01"
  name      = "my-vnet"
  parent_id = azapi_resource.resourceGroup.id
  location  = azapi_resource.resourceGroup.location

  body = {
    tags = {
      "env" = "prod"
    }
    # other prop omitted
  }

  ignore_body_changes = var.ignore_tags ? ["tags"] : []
}
```

If another tag were to be added (even outside of terraform), no diff will be produced.

This property is designed to be a fallback for `lifecycle.ignore_changes` when variables and non-static values are involved.

Fixes #1137

## Testing

- acctest: https://github.com/Azure/terraform-provider-azapi/pull/1192/checks?check_run_id=92217318793
- Manual testing


internal/services/azapi_resource_ignore_body_changes_test.go
LingyuTang, line 1: Looks like the current tests only cover the `tags` path, which has special handling in `ModifyPlan`, so they don't actually exercise the general override logic. In that case, it might be worth adding a few more scenarios to add the coverage, for example:

- Multiple/nested path such as `["properties.linkid", "properties.hello"]`
- Removing a path from `ignore_body_changes` and then adding it back
- Ignoring a path that doesn't exist in the current state/config
- Changes to both ignored and non-ignored paths in the same plan

internal/services/ignore_body_changes_private.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	return d.SetKey(ctx, pkIgnoreBodyChanges, b)
+}
+
+func (m IgnoreBodyChangesPrivateMgr) Diff(ctx context.Context, d PrivateData, paths []string) (bool, diag.Diagnostics) {
+	storedPaths, diags := m.Get(ctx, d)
+	if diags.HasError() {
+		return false, diags
+	}
+	return !slices.Equal(storedPaths, paths), diags
```
LingyuTang, line 54: This `slice.Equal` function is sensitive to ordering, which might cause `plan.Output` to be unknown and trigger reapply when user just change the order. Shall we consider sorting the slice before comparing?

LingyuTang (dismissed):
Hey Gerry, this is a really cool implementation using private state. Left some comments for discussion but overall looks good to me.

internal/services/ignore_body_changes_private.go
gerrytan: Nice catch, fixed: c7a9ec427

internal/services/azapi_resource_ignore_body_changes_test.go
gerrytan: Additional test cases added: 07288bbd7

gerrytan:
Thx @LingyuTang , all the feedbacks have been addressed. Waiting for latest acctest run to finish: https://github.com/Azure/terraform-provider-azapi/pull/1192/checks?check_run_id=92505095691

Can you please take another look at this PR meanwhile 🙏

LingyuTang (approved):
Thanks for addressing the comments! Looks all good to merge once the ACC test has passed. 👍

PR #1193, opened 2026-08-06
Fix retries for resource existence checks

### LINKED ISSUE #1179: Poll after PUT does not apply retry-logic
In our subscription vending logic, our deployment principal does not have access to all subscription resources by default. Instead, the principal creates a resource group and assigns itself a role on the resource group. The resources within the resource group then use the retry-feature to handle the Azure-internal propagation delay by retrying on "AuthorizationFailed"-errors.

So far, our deployment principal had "*/read" on the subscription level, and this worked well. However, we now wanted to harden the permissions and only add the required "/read"-permissions to the resource group-specific role. 

This means, e.g. for `Microsoft.ManagedIdentity/userAssignedIdentities`, the principal does not have any rights on the subscription-level, and it gets `Microsoft.ManagedIdentity/userAssignedIdentities/read`, `Microsoft.ManagedIdentity/userAssignedIdentities/write`, ... on the resource group-level.

However, with these reduced permissions, AzAPI often fails when polling for the resource after the PUT. It seems to me, like the retry-logic is only applied to the initial PUT-request, but not to the polling afterwards.

```
checking for presence of existing Resource: (ResourceId
"/subscriptions/{redacted}/resourceGroups/rg-test/providers/Microsoft.ManagedIdentity/userAssignedIdentities/id-test"
/ Api Version "2025-01-31-preview"): GET
https://management.azure.com/subscriptions/{redacted}/resourceGroups/rg-test/providers/Microsoft.ManagedIdentity/userAssignedIdentities/id-test
--------------------------------------------------------------------------------
RESPONSE 403: 403 Forbidden
ERROR CODE: AuthorizationFailed
--------------------------------------------------------------------------------
{
  "error": {
    "code": "AuthorizationFailed",
    "message": "The client '{redacted}' with object id '{redacted}' does not have authorization to perform action 'Microsoft.ManagedIdentity/userAssignedIdentities/read' over scope '/subscriptions/{redacted}/resourceGroups/rg-test/providers/Microsoft.ManagedIdentity/userAssignedIdentities/id-test' or the scope is invalid. If access was recently granted, please refresh your credentials."
  }
}
--------------------------------------------------------------------------------
```

hqhqhqhqhqhqhqhqhqhqhq:
## Summary
- Apply the user-configured retry policy to the pre-create resource existence check.
- Never retry the expected 404 response during that check, even when it matches a configured error regex.
- Add deterministic coverage for retrying a transient `AuthorizationFailed` response and for preserving the existing 404 behavior.

## Root cause
The failure shown in #1179 occurs on the existence-check GET before the PUT. That GET did not receive the resource retry configuration, so a transient `AuthorizationFailed` response failed immediately.

The PUT and its poller already use the configured retry options. This change is limited to the pre-create GET and preserves the safeguard introduced in #588, where a `ResourceNotFound` retry expression must not retry the expected 404 until the create timeout.

Fixes #1179

### DIFF
--- a/internal/clients/options.go
+++ b/internal/clients/options.go
@@ -133,6 +133,25 @@ func NewRetryOptionsForReadAfterCreate() (*policy.RetryOptions, *LastRetryError)
 	}, lastRetryErr
 }
 
+// NewRetryOptionsForResourceExistenceCheck applies user-configured retries while
+// allowing an expected 404 response to complete the existence check immediately.
+func NewRetryOptionsForResourceExistenceCheck(rtry retry.RetryValue) (*policy.RetryOptions, *LastRetryError) {
+	retryOptions, lastRetryErr := NewRetryOptions(rtry)
+	if retryOptions == nil {
+		return nil, nil
+	}
+
+	shouldRetry := retryOptions.ShouldRetry
+	retryOptions.ShouldRetry = func(resp *http.Response, err error) bool {
+		if resp != nil && resp.StatusCode == http.StatusNotFound {
+			return false
+		}
+		return shouldRetry(resp, err)
+	}
+
+	return retryOptions, lastRetryErr
+}
+
 // NewRetryOptions creates a RetryOptions based on the provided retry.RetryValue.
 func NewRetryOptions(rtry retry.RetryValue) (*policy.RetryOptions, *LastRetryError) {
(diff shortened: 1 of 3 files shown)


gerrytan (approved):
Thanks @hqhqhqhqhqhqhqhqhqhqhq , this PR looks good to me

PR #1196, opened 2026-08-07
fix: bicep-types-update pipeline + TF version externalisation

gerrytan:
This PR fixes the patching logic of bicep-types-az repo which causes [bicep-types-update pipeline](https://github.com/Azure/terraform-provider-azapi/actions/workflows/bicep-types-update.yaml) to fail. The latest version of bicep-types-az have removed the restriction against `list*` operations, hence it's no longer needed in our patch.

I also changed manual push pattern to allow wildcard suffix to prevent branch collision when testing and externalised the TF CLI version used by the pipelines to avoid mismatch.

## Testing

- Run the pipeline manually: https://github.com/Azure/terraform-provider-azapi/actions/runs/31138077942


gerrytan:
FYI @LingyuTang I'll send this PR for your review after the testing pipeline turn green.

LingyuTang (approved):
Thanks for fixing this issue. Looks cool to me 👍

Issue #1040, opened 2026-01-27
question: Is there a way to get the value of Key Vault secrets using this provider?


tiwood:
Is it possible (or planned) to use this provider to get the values of Key vault contents (e.g. secrets)?

With `azurerm` one has to have a provider block for the subscription where the Key vault resides in.

As a workaround there is the [azurekeyault](https://github.com/tiwood/terraform-provider-azurekeyvault) that uses the data plane API, but it would be better to use a partner provider for most scenarios.

stemaMSFT:
Closing as completed — this was resolved in v2.8.0, which added `Microsoft.KeyVault/vaults/secrets` and `Microsoft.KeyVault/vaults/keys` support via `azapi_data_plane_resource`. You can now read Key Vault secret values using the data-plane resource. Please reopen if there are any issues.

tiwood:
@stemaMSFT, maybe I'm missing something but there is no azapi_data_plane_resource *Data source* which would allow us to READ secrets.

Also it seems normal users can't reopen issues.

LingyuTang:
Hey @tiwood thanks for raising the issue. We just introduced the data source and ephemeral for data plane resource. (https://github.com/Azure/terraform-provider-azapi/pull/1199) 

You should be able to read the key vault secrets from next release. Please feel free to comment here or open a new issue if you have any question.

tiwood:
Hi @LingyuTang,

thank you, this is wonderful news, but I cannot see the data source you've added in the documentation of the current release 2.12.0, nor in the CHANGELOG itself.

Will this be released at a later point?

LingyuTang:
Hey @tiwood yes, it will be in the next release.

tiwood:
@LingyuTang should this issue be reopened until the release is complete?

LingyuTang:
Hey @tiwood, we usually close issues once the corresponding fix has been merged for clearer tracking, as the requested functionality has already been implemented and doesn't require further engineering work. If you'd like to try it before next release, feel free to clone the repository and build the provider locally.

If you find that the new implementation doesn't fully address your scenario, please leave a comment here and we'll be happy to take another look and reopen the issue if needed.

PR #1199, opened 2026-08-07
`azapi_data_plane_resource` - support data source and ephemeral resource 

LingyuTang:
## Description 
This PR add support for data source and ephemeral resource for data plane resource. 

1. New data source `azapi_data_plane_resource`, based on the PR #1141, adding test cases. 
2. New ephemeral resource `azapi_data_plane_resource`
3. The schema and `Read()`/`Open()` functions are mostly from `azapi_data_plane_resource` schema and its `read` function implementation. 

acc test: https://dev.azure.com/azclitools/internal/_build/results?buildId=336727&view=results
related issue: #1040


docs/data-sources/data_plane_resource.md
```diff
@@ -0,0 +1,117 @@
...
+This data source can read Azure data plane resources.
+
+## Schema
+
+### Required
+
+- `parent_id` (String) The ID of the azure resource in which this resource exists.
+
+	-> String length must be at least 1.
+- `type` (String) In a format like `<resource-type>@<api-version>`. `<resource-type>` is the Azure resource type, for example, `Microsoft.Storage/storageAccounts`. `<api-version>` is version of the API used to manage this azure resource.
```
gerrytan: Can we say what `type`s are supported here? Presumably they're the same as: https://registry.terraform.io/providers/Azure/azapi/latest/docs/resources/data_plane_resource#available-resources? We can just add a link there if so.

docs/data-sources/data_plane_resource.md
```diff
@@ -0,0 +1,117 @@
...
+
+	-> Ensure this in resource type format: `<resource-type>@<api-version>`.
+
+### Optional
+
+- `headers` (Map of String) A map of headers to include in the request.
```
gerrytan: Is there a hard requirement to expose this attribute? I am skeptical about allowing arbitrary header to be set from security point of view. I think we should not expose this and think about this later when the need arise.

docs/data-sources/data_plane_resource.md
```diff
@@ -0,0 +1,117 @@
...
+- `name` (String) Specifies the name (identifier segment) of the data plane resource.
+- `query_parameters` (Map of List of String) A map of query parameters to include in the request.
```
gerrytan: Similar concern to `headers`, we shouldn't expose this if possible.

internal/services/azapi_data_plane_resource_data_source.go
```diff
@@ -0,0 +1,202 @@
...
+	readTimeout, diags := model.Timeouts.Read(ctx, 5*time.Minute)
+	response.Diagnostics.Append(diags...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, readTimeout)
+	defer cancel()
+
+	id, err := parse.NewDataPlaneResourceId(model.Name.ValueString(), model.ParentID.ValueString(), model.Type.ValueString())
```
gerrytan, line 135: In the doc `name` is optional, can you please check this logic will work ok if name is not provided?

gerrytan (changes requested):
Thanks @LingyuTang , I left some question / design feedback on this.

docs/data-sources/data_plane_resource.md
LingyuTang: Added the link to the supported types to the docs and also align the existing data_plane_resource to use the same string

docs/data-sources/data_plane_resource.md
LingyuTang: Attribute removed for both data source and ephemeral

docs/data-sources/data_plane_resource.md
LingyuTang: Same as above

internal/services/azapi_data_plane_resource_data_source.go
LingyuTang: Confirmed that it will work and also added the acc test for data source that doesn't have name (singleton data plane resource)

LingyuTang:
Hey @gerrytan thanks for the review, made the changed based on our discussion. I have run the test locally and it runs okay. But will merge after the pipeline acc test pass.

gerrytan (approved):
Thanks @LingyuTang for addressing the feedbacks, this PR looks good now.

PR #1206, opened 2026-08-13
docs: Contrib guide improvements

gerrytan:
Added PR guide to limit scope of changes and split content into its own CONTRIBUTING.md file


LingyuTang (approved):
Looks cool to me. We can start with this and build upon it gradually.

PR #1212, opened 2026-08-18
feat: add support for creating dataset objects for Microsoft Foundry

jrviray:
## Description

This PR adds Azure AI Foundry dataset version support to `azapi_data_plane_resource` through a customization that orchestrates pending-upload creation, file streaming, and dataset version creation. It also updates data-plane request construction so API versions are sent as query parameters and supports explicit request content types required by the Foundry API.

### Creating Dataset Workflow

According to the product team, creating a dataset Azure asset in Foundry portal happens as follows:

1. Calling [`startPendingUpload`](https://ai.azure.com/api-reference/datasets/start-pending-upload-version/) for `/datasets/{name}/versions/{version}/startPendingUpload`.
2. Uploading the file to the returned folder using its SAS URI.
3. Calling [`create-or-update-version`](https://ai.azure.com/api-reference/datasets/create-or-update-version/) for `/datasets/{name}/versions/{version}`.

This PR reproduces that workflow in Terraform. The product group has stated that folder uploads are not currently supported through the API, so this workflow targets file-based datasets. It accepts a user-provided `source_url`. Consumers are responsible for applying their own source URL validation and host allow lists. The provider starts the pending upload, streams the source file to the returned SAS URI, and then creates or updates the dataset asset.

## Solution

The new customization translates Terraform configuration into the Foundry dataset workflow while keeping provider-only upload fields separate from the API request body. It uses the Foundry version endpoint with `PUT` and `application/json`, supports optional SHA-256 validation, exposes the computed checksum in resource output, and prevents redirect-based credential forwarding.

The data-plane client now consistently places `api-version` in the query string, preserves request query parameters and headers, and supports explicit content types for customized operations.

### 1. Foundry Dataset Customization

**File:** `internal/services/customization/foundry_dataset_customization.go`

Implements the dataset version lifecycle for `Microsoft.Foundry/datasets/versions`:

| Operation | Method | Endpoint | Notes |
|-----------|--------|----------|-------|
| **Start upload** | `POST` | `{resource}/startPendingUpload` | Requests a BlobReference upload |
| **Upload file** | `PUT` | Returned SAS URI | Streams the source file as a block blob |
| **Create/update** | `PUT` | `{resource}` | Sends the dataset version as `application/json` |
| **Read** | `GET` | `{resource}` | Retrieves the dataset version |
| **Update** | N/A | N/A | Dataset versions are immutable; a new version is required |

**Key implementation details:**

- Uses the resource `name` as the dataset version path segment.
- Sends `source_url` and `source_sha256` only through provider-side workflow logic.
- Supports optional SHA-256 verification during upload.
- Exposes `computed_sha256` in the resource output.
- Rejects redirects for both source downloads and SAS uploads.
- Redacts errors returned from dataset API operations.
- Uses a generic SHA-256 mismatch message without exposing supplied or computed values.
- Treats checksum calculation during refresh as best effort when the source URL is unavailable.
- Preserves configured body values and excludes provider-computed checksum values from the Terraform request body.

### 2. Data Plane Client Request Handling

**Files:**

- `internal/clients/data_plane_client.go`
- `internal/clients/data_plane_action_url_test.go`

Updated data-plane request construction to:

- Send `api-version` as a query parameter instead of a path segment.
- Preserve existing query parameters and headers.
- Correctly append action paths such as `/startPendingUpload`.
- Support explicit request content types for customized operations.
- Validate action URL construction independently.

### 3. Data Plane Resource Mapping

**File:** `internal/services/parse/data_plane_resources.json`

Added the dataset mapping:

- `Microsoft.Foundry/datasets/versions`
- `examples/resources/azapi_data_plane_resource/Microsoft.Foundry_datasets/main.tf`

### 4. Example Configuration

**File:** `examples/resources/azapi_data_plane_resource/Microsoft.Foundry_datasets/main.tf`

Demonstrates:

- Dataset name, version, description, type, and format configuration.
- File-based dataset uploads using `uri_file`.
- Module-level HTTPS validation for `source_url`.
- Consumer-defined host allowlist validation.
- Optional source checksum validation.
- Dataset ID and computed checksum outputs.
- Foundry API version `2025-05-01`.

### 5. Documentation

**File:** `docs/resources/data_plane_resource.md`

Regenerated the data-plane resource documentation to include the dataset mapping and updated example configuration.

### 6. Tests

**Files:**

- `internal/services/customization/foundry_dataset_customization_test.go`
- `internal/clients/data_plane_action_url_test.go`

Added coverage for:

- Pending-upload request construction.
- Upload response parsing and malformed response handling.
- Dataset version request shaping and invalid-value validation.
- Dataset defaults and generated-version rejection.
- Container and blob SAS URL handling.
- Streaming file uploads.
- SHA-256 calculation, mismatch handling, and upload suppression after validation failure.
- Upload failure handling.
- Read output checksum behavior.
- Terraform plan body handling.
- Redirect refusal for source and upload URLs.
- Redaction of sensitive API error details.
- Data-plane action URL and API-version query construction.

## Testing

The following targeted tests pass:

```text
go test ./internal/clients ./internal/services/customization ./internal/services/parse ./internal/services
```

## AI-Assisted Development Disclaimer

AI was used to assist with developing the provider implementation under the direction of user-provided requirements and input. Testing was performed to verify the proposed workflow, including request shaping, pending-upload handling, streaming uploads, checksum verification, output handling, error redaction, API-version query construction, and redirect behavior.


jrviray:

LingyuTang (commented):
Hey @jrviray, thanks for submitting the PR. We would like to support Microsoft Foundry dataset resources, but could we narrow this PR to the changes required for that feature?

1. data_plane_client.go: Instead of changing all the request function for all data plane resource, could we create individual functions that needed for this PR? A broader refactor should be a separate PR.
2. azapi_data_plane_resource: is `DataPlaneResourceIdentityModel` required for this change? If not, could we remove from this PR and submit separate PR if needed? 
3. interface.go:  Are `DataPlaneResourceWithStateBody`, `DataPlaneResourceWithPlanBody`, and all parts of `DataPlaneResourceWithReadOptions` necessary? Please remove unused hooks or introduce broader lifecycle extensions separately with focused tests.

This will reduce the regression risk and make the Foundry changes easier to review. 

Also, for the customization, could we add acc test for the new introduced data plane resource? 

Thanks, and feel free to let me know if you have any questions or want to discuss anything.

jrviray:
Thanks for the review. I narrowed the PR to the changes per your guidance

1. **`data_plane_client.go`**: Removed the broader request refactor and added the additive `ActionWithContentType` method required by the dataset workflow. Existing data-plane CRUD and action behavior remains unchanged.

2. **`azapi_data_plane_resource.go`**: Removed `DataPlaneResourceIdentityModel` and the related identity-based import handling because it was not required for dataset support.

3. **`interface.go`**: Removed `DataPlaneResourceWithStateBody` because it was unused. The remaining hooks are required for dataset behavior:

    - **`DataPlaneResourceWithPlanBody`** preserves dataset values such as `version`, `type`, and `format` during planning while keeping `computed_sha256` out of the request body.
    - **`DataPlaneResourceWithReadOptions`** preserves provider-managed fields such as `source_url` and `source_sha256`, augments read output with `computed_sha256`, and uses the augmented response as the resource output.

Please let me know if there are any other areas I need to address, thanks!

internal/services/customization/foundry_dataset_customization.go
```diff
@@ -0,0 +1,1012 @@
...
+	if err != nil {
+		return "", "", err
+	}
+
+	uploadURL, err := datasetResponseString(credential, "sasUri")
+	if err != nil {
+		return "", "", err
+	}
+
+	dataURIValues := values
```
hqhqhqhqhqhqhqhqhqhqhq, line 427: Without `blobReferenceForConsumption`, this looks for `blobUri` at the root. The [documented response](https://github.com/Azure/azure-rest-api-specs/blob/4ab86f6985d73c4fa4812a79cbdd334f47ef9837/specification/ai/data-plane/Azure.AI.Projects/stable/2025-05-01/examples/Datasets_StartPendingUploadVersion_MaximumSet_Gen.json#L16) puts it under `blobReference`, so decoding fails before the upload. Could we use that as the fallback? A regression case can reuse the existing test fixture with this field removed:

```go
delete(response, "blobReferenceForConsumption")
```

internal/services/customization/foundry_dataset_customization.go
```diff
@@ -0,0 +1,1012 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	responseBody, err := client.DataPlaneClient.ActionWithContentType(
+		ctx,
+		id.AzureResourceId,
+		"",
+		id.ApiVersion,
+		http.MethodPut,
```
hqhqhqhqhqhqhqhqhqhqhq, line 836: The example selects `2025-05-01`, whose [contract](https://github.com/Azure/azure-rest-api-specs/blob/4ab86f6985d73c4fa4812a79cbdd334f47ef9837/specification/ai/data-plane/Azure.AI.Projects/stable/2025-05-01/azure-ai-projects.json#L429) specifies PATCH with `application/merge-patch+json`. This sends PUT with `application/json`. Can we match the documented operation and assert the method and content type in a test?

internal/services/customization/foundry_dataset_customization.go
```diff
@@ -0,0 +1,1012 @@
...
+		verifySHA256,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	versionBody, _, err := datasetVersionRequestBody(
+		requestBody,
+		resourceVersion,
+		dataURI,
```
hqhqhqhqhqhqhqhqhqhqhq, line 825: With a container SAS, we append the filename for the upload but leave it out of the registered `dataUri`:

```text
Uploaded file:  .../container/data.jsonl
dataUri:        .../container
```

For `uri_file`, the registered URI needs to identify the file too. Please cover upload and registration together; the current request-body test expects the container URI. Folder datasets should keep their folder URI.

internal/services/azapi_data_plane_resource.go
```diff
@@ -328,6 +328,38 @@ func (r *DataPlaneResource) ModifyPlan(ctx context.Context, request resource.Mod
...
+				if err != nil {
+					response.Diagnostics.AddError("Invalid plan body", err.Error())
+					return
+				}
+				bodyJSON, err := json.Marshal(normalizedBody)
+				if err != nil {
+					response.Diagnostics.AddError("Invalid plan body", err.Error())
+					return
+				}
+				plan.Body, err = dynamic.FromJSONImplied(bodyJSON)
```
hqhqhqhqhqhqhqhqhqhqhq, line 354: If `source_url` comes from a changing dependency, it can be unknown during planning. This JSON round trip turns that unknown into `""`, which can make the plan inconsistent when the URL resolves at apply. Could we preserve nested unknowns and add a Terraform-value test? The current map-based helper test can't exercise this case.

internal/services/customization/foundry_dataset_customization.go
```diff
@@ -0,0 +1,1012 @@
...
+			_ []*http.Request,
+		) error {
+			return fmt.Errorf("dataset upload redirects are disabled")
+		},
+	}
+}
+
+func downloadDatasetSHA256(sourceURL string) (checksum string, err error) {
+	request, err := http.NewRequestWithContext(
+		context.Background(),
```
hqhqhqhqhqhqhqhqhqhqhq, line 518: This request uses `context.Background()` and a client without a timeout, so a stalled source can keep refresh running after the resource's read timeout. Please pass the read context through and add a test where the response body stalls until cancellation.

hqhqhqhqhqhqhqhqhqhqhq (commented):
I checked the dataset code and tests against the API docs and left some comments below.

PR #1215, opened 2026-08-18
fix: extend list unique id with nested properties

### LINKED ISSUE #1211: list_unique_id_property cannot address nested identifier fields (e.g. subnet.id), making tf plan/apply output "dirty"
## Provider / Terraform versions
* Azure/azapi v2.12.0
* Terraform 1.14.3

## Usecase:
We are a platform team providing a shared eventhub to product teams. We have split our offering into "core" (which creates the azure eventhub) and "public" (which lets teams configure topics/subscriptions) due to the teams also configuring whitelisting we are using `azapi_update_resource` from "public" to update the eventhub created in "core".

## What it doesn't work

The documentation for `list_unique_id_property` states that it exists to "avoid perpetual drift from server-side ordering". That works when the unique identifier is a **top-level string** on each list item, **BUT** a number of Azure APIs identify list items by a nested field.

A concrete example for my case is `Microsoft.EventHub/namespaces/networkRuleSets`, where each item of `properties.virtualNetworkRules` is shaped:

```json
{
  "ignoreMissingVnetServiceEndpoint": false,
  "subnet": { "id": "/subscriptions/.../subnets/example" }
}
```

There is NO top-level string field on the item at all - the identity of the rule is subnet.id. As a result there is no value of list_unique_id_property that allows the provider to match these items, and it silently falls back to positional matching. 

### Expected behaviour
Setting `list_unique_id_property = { "properties.virtualNetworkRules" = "subnet.id" } `should match list items by the nested subnet.id value, making the comparison order-insensitive.

### Actual behaviour

The setting has no effect. Items are matched positionally, so whenever the service returns the list in a different order than the one submitted, every run produces a reorder diff that apply cannot converge.

## Reproduction example:
<details><summary>A simple example to try out</summary>

```hcl
terraform {
  required_providers {
    azapi = {
      source  = "Azure/azapi"
      version = "2.12.0"
    }
    azurerm = {
      source  = "hashicorp/azurerm"
      version = "~> 4.0"
    }
  }
}

provider "azurerm" {
  features {}
}

variable "prefix" {
  type    = string
  default = "azapirepro"
}

variable "location" {
  type    = string
  default = "northeurope"
}

# Flips the order of an otherwise identical rule set.
# The desired state is unchanged: same three subnets, same settings.
variable "reverse_rule_order" {
  type    = bool
  default = false
}

resource "azurerm_resource_group" "example" {
  name     = "${var.prefix}-rg"
  location = var.location
}

resource "azurerm_virtual_network" "example" {
  name                = "${var.prefix}-vnet"
  address_space       = ["10.0.0.0/16"]
  location            = azurerm_resource_group.example.location
  resource_group_name = azurerm_resource_group.example.name
}

resource "azurerm_subnet" "example" {
  count = 3

  name                 = "subnet-${count.index}"
  resource_group_name  = azurerm_resource_group.example.name
  virtual_network_name = azurerm_virtual_network.example.name
  address_prefixes     = [cidrsubnet("10.0.0.0/16", 8, count.index)]
  service_endpoints    = ["Microsoft.EventHub"]
}

resource "azurerm_eventhub_namespace" "example" {
  name                = "${var.prefix}-ns"
  location            = azurerm_resource_group.example.location
  resource_group_name = azurerm_resource_group.example.name
  sku                 = "Standard"
}

locals {
  subnet_ids = azurerm_subnet.example[*].id

  # Same set of rules either way - only the ordering differs.
  eventhub_network_ids = var.reverse_rule_order ? reverse(local.subnet_ids) : local.subnet_ids
}

resource "azapi_update_resource" "network_rules" {
  type        = "Microsoft.EventHub/namespaces/networkRuleSets@2024-01-01"
  resource_id = "${azurerm_eventhub_namespace.example.id}/networkRuleSets/default"

  # Intended to make the list order-insensitive. Has no effect: the identity of
  # each item is the nested subnet.id, which cannot be addressed here.
  list_unique_id_property = {
    "properties.virtualNetworkRules" = "subnet.id"
  }

  body = {
    properties = {
      defaultAction       = "Deny"
      publicNetworkAccess = "Enabled"
      virtualNetworkRules = [
        for subnet_id in local.eventhub_network_ids : {
          ignoreMissingVnetServiceEndpoint = false
          subnet = {
            id = subnet_id
          }
        }
      ]
    }
  }
}
``` 

Steps
```bash
terraform apply -auto-approve
terraform plan -var reverse_rule_order=true   # same three rules, different order
```

Expected:
```bash
No changes.  — the rule set is a set of three rules; reordering it changes nothing.
```
Actual:
```
 # azapi_update_resource.network_rules will be updated in-place
  ~ resource "azapi_update_resource" "network_rules" {
      ~ body = {
          ~ properties = {
              ~ virtualNetworkRules = [
                  ~ {
                      ~ subnet = {
                          ~ id = ".../subnets/subnet-0" -> ".../subnets/subnet-2"
                        }
                    },
                    # (1 unchanged element hidden)
                  ~ {
                      ~ subnet = {
                          ~ id = ".../subnets/subnet-2" -> ".../subnets/subnet-0"
                        }
                    },
                ]
            }
        }
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```
</details> 

## Suggested fix:

Since we already have this function also doing composite keys, I recon we can just allow the identifier key to be a dot-separated path and resolve it by walking nested objects, so that in the case of my example `subnet.id` resolves to the nested string. Composite keys could continue to work by splitting on commas first and resolving each part as a path.

This would also help other Azure APIs whose list items are identified by a nested resource id.

ivan-penchev:
Hi👋 this is my first PR here, I read the contribution docs, but apologies if I missed anything.
This resolves #1211 .

I modeled it based on the original PR: #1033 which added also service tests, but I am not sure if I should add it for this or are just unit ones fine?

### DIFF
--- a/docs/resources/resource.md
+++ b/docs/resources/resource.md
@@ -107,7 +107,7 @@ output "quarantine_policy" {
 - `ignore_other_items_in_list` (List of String) A list of list property paths where items not specified in configuration should be ignored. This is intended for partial list management when combined with `list_unique_id_property` (for example, to avoid perpetual drift from server-side ordering).
 
 	-> Element value must satisfy all validations: string length must be at least 1.
-- `list_unique_id_property` (Map of String) A mapping of list property paths to the field name used as a unique identifier when comparing and merging list items. When not set, list items are matched by a `name` property (if present) or by list ordering. To match using multiple fields, specify a comma-separated list of field names (e.g., `"category, categoryGroup"`).
+- `list_unique_id_property` (Map of String) A mapping of list property paths to the field name used as a unique identifier when comparing and merging list items. When not set, list items are matched by a `name` property (if present) or by list ordering. To match using multiple fields, specify a comma-separated list of field names (e.g., `"category, categoryGroup"`). A field name may also be a dot-separated path addressing a nested field (e.g., `"subnet.id"`); a field whose own name contains dots takes precedence over the equivalent path.
 - `location` (String) The location of the Azure resource.
 
 	~> Use the state value when new value is functionally equivalent to the old and thus no change is required.
--- a/docs/resources/update_resource.md
+++ b/docs/resources/update_resource.md
@@ -107,7 +107,7 @@ resource "azapi_update_resource" "example" {
 - `ignore_other_items_in_list` (List of String) A list of list property paths where items not specified in configuration should be ignored. This is intended for partial list management when combined with `list_unique_id_property` (for example, to avoid perpetual drift from server-side ordering).
 
 	-> Element value must satisfy all validations: string length must be at least 1.
-- `list_unique_id_property` (Map of String) A mapping of list property paths to the field name used as a unique identifier when comparing and merging list items. When not set, list items are matched by a `name` property (if present) or by list ordering. To match using multiple fields, specify a comma-separated list of field names (e.g., `"category, categoryGroup"`).
+- `list_unique_id_property` (Map of String) A mapping of list property paths to the field name used as a unique identifier when comparing and merging list items. When not set, list items are matched by a `name` property (if present) or by list ordering. To match using multiple fields, specify a comma-separated list of field names (e.g., `"category, categoryGroup"`). A field name may also be a dot-separated path addressing a nested field (e.g., `"subnet.id"`); a field whose own name contains dots takes precedence over the equivalent path.
 - `locks` (List of String) A list of ARM resource IDs which are used to avoid create/modify/delete azapi resources at the same time.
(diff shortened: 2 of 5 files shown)


ivan-penchev:
Hi @gerrytan , hope you're having a good week! I know you likely have a lot on your plate, but I just wanted to gently bump this PR. Let me know if you have any initial feedback or if there is anything you need from me to get this reviewed.

PR #1217, opened 2026-08-20
feat(schema): expose resource property capabilities

jaredfholgate:
## Summary
- expose a stable public API for top-level resource property capabilities
- source results from AzAPI's embedded generated Bicep schema definitions
- cover writable, read-only, unsupported, unknown, and invalid input cases

## Validation
- go test ./pkg/schema -count=1
- go vet ./pkg/schema
- go test ./internal/azure -run Test_AllBicepTypes -count=1 -timeout 30m

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,9 @@ terraform.tfplan
 terraform.tfstate
 bin/
 modules-dev/
-/pkg/
+/pkg/*
+!/pkg/schema/
+!/pkg/schema/**
 website/.vagrant
 website/.bundle
 website/build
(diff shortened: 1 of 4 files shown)


jaredfholgate:
Apologies for this, I thought this would solve a problem for us, but it causes a coupling problem. I'm closing it for now.

jaredfholgate:
Closing this because the final architecture no longer imports or releases against AzAPI.

[Ruleset PR #155](https://github.com/Azure/tflint-ruleset-avm/pull/155) now independently generates a compact tag-capability snapshot from public `azure-rest-api-specs` and `bicep-types-az` output, embeds it for standalone linting, and updates it through a weekly review PR. This removes cross-team release coupling while preserving the same authoritative source pipeline.

The public schema API implemented here is therefore no longer required by the AVM ruleset. Thank you for the implementation and review context.

PR #1220, opened 2026-08-26
Propagate resource headers to LRO polls

### LINKED ISSUE #1218: Poll after PUT does't include headers set via `create_headers`/`update_headers`
**azapi version**: 2.7.0
**Terraform version:** 1.15.8

Headers set via `create_headers`/`update_headers` are attached to the initial PUT/PATCH request only. The poller's follow-up GETs don't include them. This breaks any endpoint that requires the header on every request for authorization.

1. PUT succeeds, header present:
```
  PUT https://<gateway>/subscriptions/.../virtualMachineScaleSets/<name>?api-version=2023-09-01
  X-Ms-Target-Resource-Id: <redacted>

  → 201 Created
  Azure-AsyncOperation: https://<gateway>/.../operations/<id>?api-version=2022-11-01&...
```

2. Poller's GET to that Azure-AsyncOperation URL — header missing:
```
  GET https://<gateway>/.../operations/<id>?api-version=2022-11-01&...
  Authorization: REDACTED
  User-Agent: ...
  X-Ms-Correlation-Request-Id: ...
  (no X-Ms-Target-Resource-Id)

  → 400 Bad Request

  {
    "type": "https://tools.ietf.org/html/rfc9110#section-15.5.1",
    "title": "One or more validation errors occurred.",
    "status": 400,
    "traceId": "..."
  }
```
3. Create ultimately fails:
```
  Error: Failed to create/update resource
  creating/updating Resource: (...): GET https://<gateway>/.../operations/<id>
  RESPONSE 400: 400 Bad Request
  ERROR CODE UNAVAILABLE
```

hqhqhqhqhqhqhqhqhqhqhq:
Fixes #1218

## Summary
- apply `create_headers` and `update_headers` through the request context so Azure SDK LRO polling requests receive them
- add a regression test covering the initial PUT, async-operation GET, and final resource GET

## Testing
- `go test ./internal/clients -count=1`
- `make test`

### DIFF
--- a/internal/clients/resource_client.go
+++ b/internal/clients/resource_client.go
@@ -46,6 +46,14 @@ func NewResourceClient(credential azcore.TokenCredential, opt *arm.ClientOptions
 }
 
 func (client *ResourceClient) CreateOrUpdate(ctx context.Context, resourceID string, apiVersion string, body interface{}, options RequestOptions) (interface{}, error) {
+	if len(options.Headers) != 0 {
+		headers := make(http.Header, len(options.Headers))
+		for key, value := range options.Headers {
+			headers.Set(key, value)
+		}
+		ctx = policy.WithHTTPHeader(ctx, headers)
+	}
+
 	// override the default retry options with the ones provided in the options
 	if options.RetryOptions != nil {
 		ctx = policy.WithRetryOptions(ctx, *options.RetryOptions)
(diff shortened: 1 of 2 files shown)


gerrytan (approved):
Thanks @hqhqhqhqhqhqhqhqhqhqhq , this PR looks good to me

PR #1221, opened 2026-08-27
azapi_resource: MoveState API version fallback

gerrytan:
- **acctest coverage**
- **azapi_resource: MoveState API version fallback**


gerrytan:
Not proceeding with this approach. Instead we'll skip the non-functional version. In the future I'll also consider introducing a provider level override so user can specify exact version to be used when moving resource from AzureRM to AzAPI.

PR #1223, opened 2026-08-28
API version index skips: added Microsoft.Portal/dashboards 2026-04-01, removed Microsoft.Keyvault 2026-02-01 and 2026-03-01-preview

### LINKED ISSUE #1216: moved from azurerm_* to azapi_resource selects an unsupported API version and drops location
### Summary

`AzapiResource.MoveState` builds the target state from the resource ID alone. It sets only `ID`, `Name`, `ParentID` and `Type`, deriving the API version from the newest embedded type rather than from anything the practitioner configured.

This produces two distinct failures when migrating an `azurerm_*` resource whose newest **embedded** API version is not served by ARM.

Unlike `terraform import`, there is **no user-facing escape hatch**: a `moved` block accepts no arguments, so the `?api-version=` workaround documented for imports (#174, #1043) cannot be applied.

### Versions

- `Azure/azapi` **v2.12.0** (latest at time of writing)
- Terraform **v1.15.8**

### Reproduction

`Microsoft.Portal/dashboards`. The provider embeds `2026-04-01`:

```
internal/azure/generated/portal/microsoft.portal/
  2015-08-01-preview  2018-10-01-preview  2019-01-01-preview
  2020-09-01-preview  2022-12-01-preview  2025-04-01-preview  2026-04-01
```

ARM does not serve `2026-04-01`:

```console
$ az provider show -n Microsoft.Portal \
    --query "resourceTypes[?resourceType=='dashboards'].apiVersions | [0]" -o tsv
2025-04-01-preview  2022-12-01-preview  2020-09-01-preview  2020-09-01-alpha
2019-01-01-preview  2018-10-01-preview  2015-08-01-preview
```

Given an existing `azurerm_portal_dashboard` in state, migrate it:

```hcl
resource "azapi_resource" "this" {
  type      = "Microsoft.Portal/dashboards@2019-01-01-preview"
  name      = "example"
  parent_id = "/subscriptions/<sub>/resourceGroups/<rg>"
  location  = "eastus"
  body      = { properties = { lenses = {} } }
}

moved {
  from = azurerm_portal_dashboard.dashboard
  to   = azapi_resource.this
}
```

### Failure 1 — plan is blocked

```
Error: Failed to retrieve resource

  with azapi_resource.this,

reading Resource: (ResourceId
"/subscriptions/.../providers/Microsoft.Portal/dashboards/example"
/ Api Version "2026-04-01"): GET https://management.azure.com/...
--------------------------------------------------------------------------------
RESPONSE 400: 400 Bad Request
ERROR CODE: NoRegisteredProviderFound
--------------------------------------------------------------------------------
{
  "error": {
    "code": "NoRegisteredProviderFound",
    "message": "No registered resource provider found for location 'eastus' and API version '2026-04-01' for type 'dashboards'. The supported api-versions are '2015-08-01-preview, 2018-10-01-preview, 2019-01-01-preview, 2020-09-01-preview, 2020-09-01-alpha, 2022-12-01-preview, 2025-04-01-preview'. ..."
  }
}
```

The configured `type` is `2019-01-01-preview`. The error names `2026-04-01`, which appears nowhere in the configuration, and nothing in the message points at `moved`, at state migration, or at the provider's embedded types.

The upgrade cannot proceed. State is not corrupted — the move is not persisted, and `terraform state list` still shows the original `azurerm_portal_dashboard` address.

### Failure 2 — `-refresh=false` silently becomes destructive

`-refresh=false` is a natural reaction to a refresh-time error, and it converts the hard failure into a destroy/recreate:

```
  # azapi_resource.this must be replaced
  # (moved from azurerm_portal_dashboard.dashboard)
-/+ resource "azapi_resource" "this" {
      + location = "eastus" # forces replacement
      ~ type     = "Microsoft.Portal/dashboards@2026-04-01" -> "Microsoft.Portal/dashboards@2019-01-01-preview"
    }

Plan: 4 to add, 0 to change, 1 to destroy.
```

`MoveState` never populates `location`, and `location` forces replacement. The post-move refresh is normally the only thing that backfills it, so suppressing refresh removes the sole guardrail.

### Root cause

https://github.com/Azure/terraform-provider-azapi/blob/main/internal/services/azapi_resource.go

```go
id, err := parse.ResourceID(azureId)

state := NewDefaultAzapiResourceModel()
state.ID       = types.StringValue(id.ID())
state.Name     = types.StringValue(id.Name)
state.ParentID = types.StringValue(id.ParentId)
state.Type     = types.StringValue(fmt.Sprintf("%s@%s", id.AzureResourceType, id.ApiVersion))
```

All four values derive from the parsed ID. An ARM resource ID encodes subscription, resource group, provider namespace, type and name — but **not** `location`, so `location` is left null. `id.ApiVersion` falls back to the newest embedded version because the `azurerm` ID carries none.

`location` cannot be inferred from the resource group either, since a resource may live in a different region than its parent RG.

### Impact

Only affects migrations of existing state; greenfield deployments are unaffected.

Severity scales with the resource type. `moved` blocks from `azurerm_*` to `azapi_resource` ship today in Azure Verified Modules for, among others, `azurerm_storage_account` and `azurerm_private_endpoint` — both location-bearing. For those, a user who works around Failure 1 with `-refresh=false` gets a destroy/recreate of a stateful resource.

Because the check is location-scoped and depends on the gap between the embedded types and ARM rollout, the same configuration can succeed in one region and fail in another, or begin failing after an `azapi` upgrade bumps the embedded default. Sovereign clouds, where API versions lag, are more exposed.

### Suggested fixes

**1. Populate `location` (and `tags`) from the source state.**
`SourceRawState` is always available and the `azurerm` state already contains them — `azurerm_portal_dashboard`, for example, declares `"location": commonschema.Location()` as required. The current `SourceSchema` only requests `id`, `resource_manager_id`, `resource_id` and `resource_versionless_id`, so these values are discarded today. This alone removes the destructive failure mode.

**2. Stop hard-selecting the newest embedded API version.**
Either fall back to the newest version ARM reports as supported, or retry the post-move read with the next-newest embedded version on `NoRegisteredProviderFound` — the approach suggested in #1089.

> **Note for implementers:** `MoveStateRequest` exposes `SourceState`, `SourceRawState`, `SourceTypeName` and `SourcePrivate`, but **not** the target resource's configuration. So "use the configured `type`'s API version" is not implementable inside `MoveState`; fix 2 needs to work without it.

Regenerating the embedded types will not help here — the embedded version is *ahead* of the deployed service rather than stale, which distinguishes this from #1043 and #1089.

### Related

- #174 — `import` ignores the API version in `type`; mitigated by `?api-version=`
- #1043 — `import` uses an unsupported API version
- #1089 — same, with a suggestion to fall back on `NoRegisteredProviderFound`

All three concern the `import` path, where `?api-version=` provides a workaround. No equivalent exists for `moved`.

gerrytan:
When moving Microsoft.Portal/dashboards from AzureRM to AzAPI, AzAPI MoveState function has no visibility of the intended API version, so it choses the latest indexed one 2026-02-01 which happen to not yet function (service team haven't completed deployment but already publishes the spec).

To address this the offending version is skipped in the hardcoded skip map.

While doing this I also notice there are outdated keyvault skips which is now functional, so they're removed.

## Testing

1. Added non-core acctest to reproduce the problem, tested working locally
2. Checked using az rest CLI the keyvault versions are now operational
3. Check for regression in acctest: https://github.com/Azure/terraform-provider-azapi/pull/1223/checks?check_run_id=98750380372

Fixes #1216

### DIFF
--- a/docs/resources/Microsoft.KeyVault_vaults_keys.md
+++ b/docs/resources/Microsoft.KeyVault_vaults_keys.md
@@ -139,7 +139,7 @@ resource "azapi_resource_action" "put_key" {
 
 The following arguments are supported:
 
-* `type` - (Required) The type of the resource. This should be set to `Microsoft.KeyVault/vaults/keys@api-version`. The available api-versions for this resource are: [`2019-09-01`, `2020-04-01-preview`, `2021-04-01-preview`, `2021-06-01-preview`, `2021-10-01`, `2021-11-01-preview`, `2022-02-01-preview`, `2022-07-01`, `2022-11-01`, `2023-02-01`, `2023-07-01`, `2024-04-01-preview`, `2024-11-01`, `2024-12-01-preview`, `2025-05-01`].
+* `type` - (Required) The type of the resource. This should be set to `Microsoft.KeyVault/vaults/keys@api-version`. The available api-versions for this resource are: [`2019-09-01`, `2020-04-01-preview`, `2021-04-01-preview`, `2021-06-01-preview`, `2021-10-01`, `2021-11-01-preview`, `2022-02-01-preview`, `2022-07-01`, `2022-11-01`, `2023-02-01`, `2023-07-01`, `2024-04-01-preview`, `2024-11-01`, `2024-12-01-preview`, `2025-05-01`, `2026-02-01`, `2026-03-01-preview`].
 
 * `parent_id` - (Required) The ID of the azure resource in which this resource is created. The allowed values are:  
   `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.KeyVault/vaults/{resourceName}`
@@ -157,5 +157,5 @@ For other arguments, please refer to the [azapi_resource](https://registry.terra
  terraform import azapi_resource.example /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.KeyVault/vaults/{resourceName}/keys/{resourceName}
  
  # It also supports specifying API version by using the resource id with api-version as a query parameter, e.g.
- terraform import azapi_resource.example /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.KeyVault/vaults/{resourceName}/keys/{resourceName}?api-version=2025-05-01
+ terraform import azapi_resource.example /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.KeyVault/vaults/{resourceName}/keys/{resourceName}?api-version=2026-03-01-preview
  ```
--- a/docs/resources/Microsoft.KeyVault_vaults_secrets.md
+++ b/docs/resources/Microsoft.KeyVault_vaults_secrets.md
@@ -104,7 +104,7 @@ resource "azapi_resource_action" "put_secret" {
 
 The following arguments are supported:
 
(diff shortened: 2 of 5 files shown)


LingyuTang (approved):
Thanks Gerry looks cool to me. Portal team was aiming to release this new version at this quarter, so we might check again in few month.

PR #1228, opened 2026-09-02
feat: add support for creating evaluations and evaluation runs for Microsoft Foundry

jrviray:
## Description

This PR adds Azure AI Foundry evaluation and evaluation run support to `azapi_data_plane_resource` through customizations for POST-to-collection APIs that return service-generated resource identifiers.

The implementation preserves existing data-plane client behavior by adding a dedicated action path for Foundry endpoints whose API version is encoded in the `/openai/v1` URL path rather than sent as an `api-version` query parameter.

### Creating an Evaluation Workflow

Foundry evaluations use a collection-based workflow:

1. Send a `POST` request to `/openai/v1/evals`.
2. Extract the generated evaluation ID from the response.
3. Use the generated ID for subsequent read, update, and delete operations.

Evaluation runs follow a similar workflow:

1. Send a `POST` request to `/openai/v1/evals/{evaluation_id}/runs`.
2. Extract the generated run ID from the response.
3. Use the generated evaluation and run IDs for subsequent read and delete operations.

These APIs do not accept a user-provided top-level resource name. Evaluation runs also require the parent evaluation ID to construct their request path.

## Solution

The new customizations use the existing create-result extension pattern to support server-generated identifiers while preserving the standard Terraform resource lifecycle.

- Evaluation and evaluation-run resources omit the top-level `name` argument.
- The service-generated identifier is extracted from the create response and stored in Terraform state.
- Evaluation runs expose a top-level `evaluation_id` argument that is used only to construct the request URL.
- `name` and `evaluation_id` validation is driven by the resource type's URL format (`data_plane_resources.json`) rather than hardcoded resource-type strings: `evaluation_id` is required when the type's path contains an `{evaluationId}` segment, and rejected when it doesn't — mirroring the existing `name`/`HasNameSegment` pattern from #1053.
- Provider-only fields are excluded from request bodies where necessary.
- Foundry `/openai/v1` requests do not receive an `api-version` query parameter.
- Existing data-plane resource request behavior remains unchanged for all other resources.

### 1. Foundry Evaluation Customization

**File:** `internal/services/customization/foundry_evaluation_customization.go`

Implements the lifecycle for `Microsoft.Foundry/evaluation/versions`:

| Operation | Method | Endpoint | Notes |
|-----------|--------|----------|-------|
| **Create** | `POST` | `{parent}/openai/v1/evals` | Extracts the service-generated evaluation ID |
| **Read** | `GET` | `{parent}/openai/v1/evals/{evaluationId}` | Retrieves the evaluation |
| **Update** | `POST` | `{parent}/openai/v1/evals/{evaluationId}` | Sends mutable evaluation metadata |
| **Delete** | `DELETE` | `{parent}/openai/v1/evals/{evaluationId}` | Deletes the evaluation |

Key implementation details:

- Uses `CreateResultFunc` to capture the generated evaluation ID.
- Validates that the create response contains a non-empty string `id`.
- Limits update requests to supported mutable fields such as `name` and `metadata`.
- Uses the Foundry-specific action path without an `api-version` query parameter.

### 2. Foundry Evaluation Run Customization

**File:** `internal/services/customization/foundry_evaluation_run_customization.go`

Implements the lifecycle for `Microsoft.Foundry/evaluation/runs`:

| Operation | Method | Endpoint | Notes |
|-----------|--------|----------|-------|
| **Create** | `POST` | `{parent}/openai/v1/evals/{evaluationId}/runs` | Extracts the service-generated run ID |
| **Read** | `GET` | `{parent}/openai/v1/evals/{evaluationId}/runs/{runId}` | Retrieves the evaluation run |
| **Update** | N/A | N/A | Evaluation runs are immutable |
| **Delete** | `DELETE` | `{parent}/openai/v1/evals/{evaluationId}/runs/{runId}` | Deletes the evaluation run |

Key implementation details:

- Requires the top-level `evaluation_id` argument; also rejects the `__generated__` sentinel value used internally for service-generated names.
- Uses `evaluation_id` to construct the request path without including it in the request body.
- Supports generated run IDs returned as `id`, `run_id`, `runId`, or `runID`.
- Rejects update operations with an explicit immutability error.
- Uses the Foundry-specific action path without an `api-version` query parameter.

### 3. Additive Data Plane Client Support

**Files:**

- `internal/clients/data_plane_client.go`
- `internal/clients/data_plane_client_test.go`

Added `ActionWithoutAPIVersion` for data-plane APIs whose version is part of the endpoint path.

The existing `Action`, `buildRequest`, `CreateOrUpdateThenPoll`, `Get`, and `DeleteThenPoll` behavior remains unchanged. Existing data-plane resources continue to use the original API-version request path.

### 4. Resource Model and Resource ID Mapping

**Files:**

- `internal/services/azapi_data_plane_resource.go`
- `internal/services/parse/data_plane_resource_id.go`
- `internal/services/parse/data_plane_resources.json`
- `internal/services/parse/data_plane_type_map.go`
- `internal/services/customization/registration.go`

Added:

- The optional `evaluation_id` resource argument.
- A data-driven `parse.HasEvaluationIdSegment` helper (alongside the existing `HasNameSegment`) that inspects each resource type's URL format to determine whether it uses an evaluation ID segment.
- Plan-time and create-time validation requiring `evaluation_id` for resource types that use it, and rejecting it for resource types that don't.
- Generated-name validation for evaluation and evaluation-run resources.
- Evaluation ID parsing and reconstruction for imports and refresh operations.
- Resource mappings for:
  - `Microsoft.Foundry/evaluation/versions`
  - `Microsoft.Foundry/evaluation/runs`

### 5. Example Configuration

**Files:**

- `examples/resources/azapi_data_plane_resource/Microsoft.Foundry_evaluation/main.tf`
- `examples/resources/azapi_data_plane_resource/Microsoft.Foundry_evaluation_run/main.tf`

Examples demonstrate:

- Creating an evaluation without a top-level `name`.
- Creating an evaluation run with `evaluation_id`.
- Configuring evaluation data sources and testing criteria.
- Configuring evaluation run targets and input message templates.
- Using Foundry API version `2025-05-01`.

### 6. Documentation

**Files:**

- `docs/resources/data_plane_resource.md`
- `internal/tools/gendoc/dataplane.go`
- `internal/tools/gendoc/dataplane_test.go`

Regenerated the data-plane resource documentation to include both Foundry resource mappings and examples.

The documentation generator now safely preserves Terraform expressions such as `{{item.input}}` when embedding examples in Go templates.

### 7. Tests

**Files:**

- `internal/services/customization/foundry_evaluation_customization_test.go`
- `internal/services/customization/foundry_evaluation_run_customization_test.go`
- `internal/clients/data_plane_client_test.go`
- `internal/services/parse/data_plane_resource_id_test.go`
- `internal/services/parse/data_plane_type_map_test.go`
- `internal/services/azapi_data_plane_resource_name_validation_test.go`
- `internal/services/azapi_data_plane_resource_import_test.go`
- `internal/tools/gendoc/dataplane_test.go`

Added coverage for:

- Evaluation and evaluation-run create request construction.
- Generated evaluation and run ID extraction.
- Foundry endpoint paths and API-version omission.
- Evaluation-run `evaluation_id` validation and body filtering.
- Symmetric `evaluation_id` validation (required vs. rejected) driven by resource-type URL format.
- Immutable evaluation-run update behavior.
- Evaluation and evaluation-run resource ID parsing and round-tripping.
- Import state handling.
- Generated-name validation.
- Legacy and Foundry-specific data-plane request builders.
- Documentation generation with Terraform template expressions.

## Testing

```text
go test ./...
```

## AI-Assisted Development Disclaimer

AI was used to assist with developing the provider implementation under the direction of user-provided requirements and input. Testing was performed to verify request construction, generated identifier handling, evaluation-run parent ID handling, resource ID parsing, import behavior, validation, documentation generation, and preservation of existing data-plane client behavior.

### DIFF
--- a/internal/services/customization/foundry_evaluation_customization.go
+++ b/internal/services/customization/foundry_evaluation_customization.go
@@ -80,11 +80,10 @@ func (c FoundryEvaluationCustomization) CreateResultFunc() CreateResultFunc {
 		body interface{},
 		options clients.RequestOptions,
 	) (parse.DataPlaneResourceId, interface{}, error) {
-		responseBody, err := client.DataPlaneClient.Action(
+		responseBody, err := client.DataPlaneClient.ActionWithoutAPIVersion(
 			ctx,
 			foundryEvaluationCollectionID(id.ParentId),
 			"",
-			"", // /openai/v1 endpoints must not receive api-version
 			http.MethodPost,
 			body,
 			options,
@@ -118,11 +117,10 @@ func (c FoundryEvaluationCustomization) ReadFunc() ReadFunc {
 		id parse.DataPlaneResourceId,
 		options clients.RequestOptions,
 	) (interface{}, error) {
-		return client.DataPlaneClient.Action(
+		return client.DataPlaneClient.ActionWithoutAPIVersion(
 			ctx,
 			id.AzureResourceId,
 			"",
-			"", // /openai/v1 endpoints must not receive api-version
 			http.MethodGet,
 			nil,
 			options,
@@ -143,11 +141,10 @@ func (c FoundryEvaluationCustomization) UpdateFunc() UpdateFunc {
 			return err
 		}
 
-		_, err = client.DataPlaneClient.Action(
+		_, err = client.DataPlaneClient.ActionWithoutAPIVersion(
 			ctx,
 			id.AzureResourceId,
 			"",
-			"", // /openai/v1 endpoints must not receive api-version
 			http.MethodPost,
 			requestBody,
 			options,
@@ -163,11 +160,10 @@ func (c FoundryEvaluationCustomization) DeleteFunc() DeleteFunc {
 		id parse.DataPlaneResourceId,
 		options clients.RequestOptions,
 	) error {
-		_, err := client.DataPlaneClient.Action(
+		_, err := client.DataPlaneClient.ActionWithoutAPIVersion(
 			ctx,
 			id.AzureResourceId,
 			"",
-			"", // /openai/v1 endpoints must not receive api-version
 			http.MethodDelete,
 			nil,
 			options,
--- a/internal/services/customization/foundry_evaluation_customization.go
+++ b/internal/services/customization/foundry_evaluation_customization.go
@@ -1,0 +1,180 @@
+package customization
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/Azure/terraform-provider-azapi/internal/clients"
+	"github.com/Azure/terraform-provider-azapi/internal/services/parse"
+)
+
+// FoundryEvaluationCustomization supports Foundry evaluations, which use a
+// collection POST for creation and a generated evaluation ID for later calls.
+//
+// Create: POST {endpoint}/openai/v1/evals
+// Get:    GET  {endpoint}/openai/v1/evals/{evalID}
+// Update: POST {endpoint}/openai/v1/evals/{evalID}
+// Delete: DELETE {endpoint}/openai/v1/evals/{evalID}
+type FoundryEvaluationCustomization struct{}
+
+const foundryEvaluationCollectionPath = "/openai/v1/evals"
+
+func (c FoundryEvaluationCustomization) GetResourceType() string {
+	return "Microsoft.Foundry/evaluation/versions"
+}
+
+func (c FoundryEvaluationCustomization) CreateFunc() CreateFunc {
+	return nil
+}
+
+func foundryEvaluationCollectionID(parentID string) string {
+	return strings.TrimRight(strings.TrimSpace(parentID), "/") + foundryEvaluationCollectionPath
+}
+
+func foundryResourceID(response interface{}) (string, error) {
+	values, err := foundryEvaluationBodyMap(response)
+	if err != nil {
+		return "", fmt.Errorf("invalid evaluation response: %w", err)
+	}
+
+	value, ok := values["id"]
+	if !ok {
+		return "", fmt.Errorf(`evaluation response field "id" is missing`)
+	}
+	evaluationID, ok := value.(string)
+	if !ok {
+		return "", fmt.Errorf(`evaluation response field "id" must be a string`)
+	}
+	if strings.TrimSpace(evaluationID) == "" {
+		return "", fmt.Errorf(`evaluation response field "id" must be a non-empty string`)
+	}
+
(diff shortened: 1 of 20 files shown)


jrviray:

internal/services/customization/foundry_evaluation_customization.go
```diff
@@ -0,0 +1,176 @@
...
+func foundryEvaluationUpdateRequestBody(body interface{}) (map[string]interface{}, error) {
+	values, err := foundryEvaluationBodyMap(body)
+	if err != nil {
+		return nil, err
+	}
+
+	// The update API only accepts mutable evaluation metadata. The data source
+	// and testing criteria are create-time properties.
+	updateBody := make(map[string]interface{}, 2)
+	for _, fieldName := range []string{"name", "metadata"} {
```
hqhqhqhqhqhqhqhqhqhqhq, line 66: Changing only `data_source_config.include_sample_schema` from `false` to `true` would be dropped here, but the provider still saves the new body to state. We shouldn't report that as applied. Could changes to `data_source_config` and `testing_criteria` require replacement, or return an error? Please test this without `replace_triggers_refs`; the existing test only checks that these fields are filtered out.

internal/services/customization/registration.go
```diff
@@ -13,6 +13,12 @@ func init() {
 
 	var foundryAgentCustomization DataPlaneResource = FoundryAgentCustomization{}
 	customizations[strings.ToLower(foundryAgentCustomization.GetResourceType())] = foundryAgentCustomization
+
+	var foundryEvaluationCustomization DataPlaneResource = FoundryEvaluationCustomization{}
+	customizations[strings.ToLower(foundryEvaluationCustomization.GetResourceType())] = foundryEvaluationCustomization
```
hqhqhqhqhqhqhqhqhqhqhq, line 18: A data source with `name = "eval_123"` hits the ["name should not be set" validation](https://github.com/Azure/terraform-provider-azapi/blob/53d6c6389470fb00e6417d999dc994722813a3be/internal/services/azapi_data_plane_resource.go#L646), even though it needs that ID to read the existing evaluation. Could we separate lookup validation from creation validation? Run lookups also need the parent evaluation ID, which the data source doesn't currently pass. Please cover both with data-source read tests.

hqhqhqhqhqhqhqhqhqhqhq (commented):
I found two issues with updates and data-source reads in the code and tests. I haven't run these cases against a live Foundry project.

Issue #988, opened 2025-09-27
feature request: support for terraform actions


jaredfholgate:
I'd like to see the azapi_resource_action re-implemented as an action as released in Terraform CLI 1.4: https://developer.hashicorp.com/terraform/language/v1.14.x/invoke-actions

Also, if we could have a generic action resource that can call any data plane API, that would be amazing.

CC: @matt-FFFFFF

ms-henglu:
Hi @jaredfholgate ,

Thank you for taking time to open this issue.

I'll work on this feature and the target release is v2.8.0. Thanks!

matt-FFFFFF:
Hi @ms-henglu this is great news thanks!

For the general purpose action - could #918 be a source if inspiration? That is a general client that can get a token for any Entra-backed API?

This would provide a general-purpose approach that meets and exceeds the capabilities in the aws provider. You will see that the [`aws_lambda_invoke`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/actions/lambda_invoke) action uses `jsonencode()`... Perhaps they will lean how to use an HCL dynamic in the future ;-)

stemaMSFT:
@matt-FFFFFF how low-level/generic do you want this data plane API? Our longstanding hesitation around this feature is that it'll end up being essentially a REST API due to the lack of consistency in how data plane APIs behave.
On the other hand, fully agree that we should be supporting actions with our next release.

stemaMSFT:
Thanks @matt-FFFFFF let me chat with the team about this. I hadn't thought about these unique `azapi` based benefits applying to a "restful" experience, but that resonates with me.

gerrytan:
I will look into adding data plane action support with following use cases as initial rollout:

1. Adding key to key vault
2. Adding blob to storage account
3. Turning off VM
4. Triggering container registry task
5. Triggering container app jobs

PR #1229, opened 2026-09-02
azapi_data_plane_resource: apply retry logic to existence check

gerrytan:
Apply retry logic to data plane resource existence check to fix following bug:

1. Create an RBAC based key vault
2. Assign "key vault crypto officer" role into current principal
3. Add a key to the keyvault using the data_plane_resource with retry config set for "Forbidden" error regex

Step 2 require 20-30 seconds to be effective despite the successful assignment confirmation returned by the API. The expectation was the retry config on step 3 will activate to poll until the role assignment become effective. However it failed with 403 right away because the resource existence check did not have retry logic applied to it.

Related to: #988

## Testing

1. Check no regression on acctest: https://github.com/Azure/terraform-provider-azapi/pull/1229/checks?check_run_id=100243295549
2. Added unit test to simulate the role assignment lag


LingyuTang (approved):
Thanks Gerry, nice fix. This looks cool to me.

PR #1233, opened 2026-09-03
Add repository-owned X agent policy

a0x1ab:
## Summary
- add the trusted `.x/x.yml` manifest with an explicit least privilege skill allowlist
- define repository owned coordinator, requirements and analyzer policies
- reserve `.x/skills/` for one function per file custom skills

## Validation
- loaded every approved agent through the central trusted policy loader

### DIFF
--- a/.x/coordinator.md
+++ b/.x/coordinator.md
@@ -1,0 +1,24 @@
+# AzAPI X Engineering Agent Coordinator
+
+Act only on `Azure/terraform-provider-azapi`. The trusted base branch is
+`main`. This is an analysis-only workflow: never assign a coding agent, create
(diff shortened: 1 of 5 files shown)


a0x1ab:
@gerrytan This trusted .x policy is green and ready for human review; the central rollout will remain fail-closed until this repository policy is approved and merged.

gerrytan (changes requested):
Thanks @a0x1ab , but AzAPI is an open source software, and I prefer it to be agent agnostic.

Can you please refactor the skill and instructions to .agent or AGENTS.md in a way that it will work regardless which agent / framework is being used.

